PR #18, opened 2018-04-23
Feature/atr 21 refactoring

SENya1990:
Tested on both VS 2015 and VS 2017


src/Acuminator/Acuminator.Analyzers/Analyzers/BQL/BqlParameterMismatchAnalyzer.cs
```diff
@@ -0,0 +1,89 @@
+﻿using System;
```
dropsonic: I suggest to remove commented code

SENya1990: This is commented diagnostic which I'll use in the future, so it will be in use, just a bit later

SENya1990: By the way, unit tests run succesfully

PR #25, opened 2018-06-18
Feature/atr 17 master pxdb field attribute not matching property

SENya1990:


SENya1990:
@k-s-popova please verify the following messages:
For PX1021 diagnostic - DAC property type doesn't correspond to the bound/unbound field attribute:
"The PXDBFieldAttribute type and property type do not correspond to each other"  - for the diagnostic
"Make the property type correspond to the attribute type."  - for the suggested code fix

For PX1023 diagnostic - multiple bound/unbound field attribute on DAC property:
"The DAC property can not be marked with multiple field attributes" - for diagnostic
"Remove other field attributes from the property." - for the code fix

SENya1990:
@k-s-popova I'll change messages, thanks

SENya1990:
@k-s-popova first message probably should not specify concrete attribute type:
"The type of DAC field attribute does not correspond to the property type"
Please verify

k-s-popova:
@SENya1990, "The type of the DAC field attribute does not correspond to the property type"

SENya1990:
@k-s-popova "Remove unnecessary field attributes from the property" message doesn't specify that the selected attribute will remain on the property. Shoul we specify it?

k-s-popova:
@SENya1990, let's change as follows:
"Leave the selected field attribute and remove other field attributes from the property"

SENya1990:
@k-s-popova Thanks!

PR #34, opened 2018-06-23
Feature/atr 143 master dac field must be abstract

SENya1990:


SENya1990:
@k-s-popova Please verify the following messages for diagnostic and code fix:
Diagnostic: "The DAC field type must be abstract"
Code Fix: "Make the DAC field type abstract"

PR #52, opened 2018-08-03
Feature/atr 73 master add bql format to context menu

SENya1990:
Merge whenever you want - created from master


src/Acuminator/Acuminator.Vsix/packages.config
```diff
@@ -30,6 +31,7 @@
   <package id="Microsoft.VisualStudio.TextManager.Interop" version="7.10.6070" targetFramework="net461" />
   <package id="Microsoft.VisualStudio.TextManager.Interop.8.0" version="8.0.50727" targetFramework="net461" />
   <package id="Microsoft.VisualStudio.Threading" version="14.1.111" targetFramework="net461" />
+  <package id="Microsoft.VisualStudio.Threading.Analyzers" version="15.8.122" targetFramework="net461" />
```
dropsonic: Is it compatible with VS2015?

src/Acuminator/Acuminator.Vsix/Formatter/FormatBqlCommand.cs
```diff
@@ -24,7 +26,7 @@ internal sealed class FormatBqlCommand
 		/// <summary>
 		/// Command ID.
 		/// </summary>
-		public const int CommandId = 0x0100;
+		public const int CommandId = 0x0101;
```
dropsonic: Why is that?

src/Acuminator/Acuminator.Vsix/Help Examples/ExampleOfSubMenuDefinition.vsct
```diff
@@ -0,0 +1,99 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
dropsonic: Do we need this file?

SENya1990: I thought it could be good as example of vcst definition. There are links to other examples too. Maybe we don't need it though.

src/Acuminator/Acuminator.Vsix/Formatter/FormatBqlCommand.cs
SENya1990: I took it from the example with submenu where submenu already had that ID. So my submenu has ID 0x0100  and its commands has IDs starting from 0x0101.

src/Acuminator/Acuminator.Vsix/packages.config
SENya1990: I haven't checked yet, I'll try it now. It is actually from ExtensibilityTolls extension I installed.

SENya1990: As far as I can see, it is compatible with VS 2015. This package contains analyzers for wrok with VS SDK, some are very useful - thanks to them I added some unobvious checks to code

src/Acuminator/Acuminator.Vsix/Help Examples/ExampleOfSubMenuDefinition.vsct
SENya1990: @dropsonic what do you think?

PR #60, opened 2018-08-09
Bugfix/atr 170 master

SENya1990:


src/Acuminator/Acuminator.Vsix/AcuminatorVSPackage.vsct
```diff
@@ -74,11 +73,11 @@
...
 				<Icon guid="guidImages" id="acumaticaIcon" />
 				<CommandFlag>DefaultDisabled</CommandFlag>
-				<CommandFlag>DefaultInvisible</CommandFlag>
-				<CommandFlag>DynamicVisibility</CommandFlag>
+				<!--<CommandFlag>DefaultInvisible</CommandFlag>
+					  <CommandFlag>DynamicVisibility</CommandFlag>-->
 				<Strings>
 					<CommandName>cmdidGoToDeclarationOrHandlerCommand</CommandName>
-					<ButtonText>Go to action/view declaration/handler</ButtonText>
+					<ButtonText>Go To Action/View Declaration/Handler</ButtonText>
```
SENya1990: @k-s-popova, please review the text for the navigation command:
"Go To Action/View Declaration/Handler"

k-s-popova: @SENya1990, can we show the message depending on where we are actually going to navigate?
That is, "Go To Action", "Go To View Declaration", "Go To Handler". It would be much more user-friendly:)
If not, we can use this message as is.

SENya1990: @k-s-popova I agree, that was the first thing that came to my mind - to generate command text dynamically. But I'm not sure we can do it without performing the analysis of the entire document on context menu opening and this will probably slow down context menu, hang the UI and become an awful experience for user.
So, I think we need some sensible generic name. Or maybe we can go with fancy command names like "AcuGoTo".

k-s-popova: @SENya1990, AcuGoTo says nothing, therefore I prefer the first variant.

PR #64, opened 2018-08-14
ATR-13: added simple logging

SENya1990:


src/Acuminator/Acuminator.Vsix/Utils/AcuminatorLogger.cs
```diff
@@ -0,0 +1,88 @@
...
+						  .AppendLine($"STACK TRACE: {e.Exception.StackTrace}")
+						  .AppendLine($"TARGET SITE: {e.Exception.TargetSite}")
+						  .AppendLine($"SOURCE: {e.Exception.Source}");
+
+			return messageBuilder.ToString();
+		}
+
+		public void Dispose()
+		{
+			if (Interlocked.CompareExchange(ref _disposed, DISPOSED, NOT_DISPOSED) == NOT_DISPOSED)
```
dropsonic: Does it mean that the package can be disposed from several threads at once?

src/Acuminator/Acuminator.Vsix/Utils/AcuminatorLogger.cs
```diff
@@ -0,0 +1,88 @@
...
+			if (activeTextView == null)
+				return;
+
+			Document currentDocument = activeTextView.TextSnapshot.GetOpenDocumentInCurrentContextWithChanges();
+
+			if (currentDocument == null)
+				return;
+
+			string logMessage = CreateLogMessage(e, currentDocument);
+			string logPath = ActivityLog.LogFilePath;
```
dropsonic: Where is it used?

src/Acuminator/Acuminator.Vsix/Utils/AcuminatorLogger.cs
```diff
@@ -0,0 +1,88 @@
...
+			ActivityLog.TryLogError(PackageName, logMessage);
```
dropsonic: [This article](https://docs.microsoft.com/en-us/visualstudio/extensibility/how-to-use-the-activity-log) states that Activity Log instance should be obtained as a service inside the `Initialize` method in the VS Package.

src/Acuminator/Acuminator.Vsix/AcuminatorVSPackage.cs
```diff
@@ -95,6 +99,12 @@ public GeneralOptionsPage GeneralOptionsPage
             }
         }
 
+		internal AcuminatorLogger AcuminatorLogger
```
dropsonic: I think we should consider a possibility of using MEF and Services to register & obtain a logger instance ([one](https://docs.microsoft.com/en-us/visualstudio/extensibility/how-to-get-a-service), [two](https://docs.microsoft.com/en-us/visualstudio/extensibility/how-to-provide-a-service), [three](https://www.c-sharpcorner.com/UploadFile/20c06b/using-mef-to-setup-dependency-injection/))

src/Acuminator/Acuminator.Vsix/Utils/AcuminatorLogger.cs
SENya1990: @dropsonic , leftover from debug,, I'll delete it

src/Acuminator/Acuminator.Vsix/Utils/AcuminatorLogger.cs
SENya1990: @dropsonic, I don't know for sure wht will happen in VS so it's better to handle this situation. Besides, the class could be used somewhere else besides the Package

PR #71, opened 2018-08-24
Feature/atr 181 master method call analysis

dropsonic:
@SENya1990 messages have already been verified by @k-s-popova in Skype


SENya1990:
@dropsonic the nested walker should check ConditionalMemberAccess too

dropsonic:
@SENya1990 I've added support for conditional access expressions

PR #84, opened 2018-08-31
Bugfix/atr 192

metallmonster:


src/Acuminator/Acuminator.Analyzers/StaticAnalysis/MethodsUsageInDac/MethodsUsageInDacAnalyzer.cs
```diff
@@ -46,29 +50,53 @@ private void AnalyzeDacDeclaration(SyntaxNodeAnalysisContext syntaxContext, PXCo
...
 			}
 		}
 
-		private void AnalyzeMethodInvocationInDacProperty(PropertyDeclarationSyntax property,
-			SyntaxNodeAnalysisContext syntaxContext, PXContext pxContext)
+        private IEnumerable<INamedTypeSymbol> GetWhitelist(PXContext pxContext)
+        {
+            return new[]
+            {
+                pxContext.SystemTypes.Bool,
```
SENya1990: @metallmonster meybe we should add enums and float here

metallmonster: @SENya1990 I looked into our code, we have PXFloatAttribute, so I'll add float
Enums aren't analyzed by this diagnostic

SENya1990: @metallmonster float should be added just in case - it's strange to not support one of the primitive types. 
Will the diagnostic show error on this code:
```
public enum MyEnum { Val };

public class DAC : IBqlTable
{
    public string Property 
    {
       get
       {
           return Val.ToString();
        }
     }
}
```

metallmonster: @SENya1990 yes, it will. Now it is fixed.

PR #86, opened 2018-09-03
Feature/atr 214 master event handlers semantic engine

dropsonic:


src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/CodeResolvingUtils.cs
```diff
@@ -377,5 +377,77 @@ public static bool IsPXSetupBqlCommand(this ITypeSymbol bqlTypeSymbol, PXContext
...
+
+				// Old syntax
+				if (symbol.Parameters[0].Type.OriginalDefinition.InheritsFromOrEquals(pxContext.PXCacheType))
+				{
+					if (symbol.Name == "CacheAttached")
+						return EventType.CacheAttached;
+
+					if (symbol.Parameters.Length >= 2)
+					{
+						switch (symbol.Parameters[1].Type.OriginalDefinition)
```
metallmonster: Depth indentation is bigger than 3. It is a violation of our code style)

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/EventHandlers/EventHandlerAnalyzer.cs
```diff
@@ -0,0 +1,62 @@
...
+namespace Acuminator.Analyzers.StaticAnalysis.EventHandlers
+{
+	[DiagnosticAnalyzer(LanguageNames.CSharp)]
+	public class EventHandlerAnalyzer : PXDiagnosticAnalyzer
+	{
+		private readonly ImmutableArray<IEventHandlerAnalyzer> _innerAnalyzers;
+
+		public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; }
+
+		public EventHandlerAnalyzer() : this(
```
metallmonster: Do we need this constructor? It looks like it can be removed.

dropsonic: It is used by Roslyn engine to create a diagnostic.

src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/CodeResolvingUtils.cs
dropsonic: It is exactly 3 levels (3 nested `if`s to avoid two `return EventType.None` statements).

metallmonster: 3 nested if + switch. Do we consider that switch adds indentation?

src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/CodeResolvingUtils.cs
```diff
@@ -377,5 +377,77 @@ public static bool IsPXSetupBqlCommand(this ITypeSymbol bqlTypeSymbol, PXContext
...
+
+		/// <summary>
+		/// Returns event handler type for the provided method symbol.
+		/// </summary>
+		/// <param name="symbol">Method symbol for the event handler</param>
+		/// <param name="pxContext">PXContext instance</param>
+		/// <returns>Event Type (e.g. RowSelecting). If method is not an event handler, returns <code>EventType.None</code>.</returns>
+		public static EventType GetEventHandlerType(this IMethodSymbol symbol, PXContext pxContext)
+		{
+			if (symbol.ReturnsVoid && symbol.TypeParameters.IsEmpty && !symbol.Parameters.IsEmpty)
```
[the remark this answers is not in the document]
dropsonic: @SENya1990 no, we discussed it before.
Event handlers can be declared in many different places outside of PXGraph / PXGraphExtension, such as attributes, PXSelectBase-derived classes (indexers), helper classes, etc.
When there is a need for that flag, we will add it. Right now it is not needed.

src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/CodeResolvingUtils.cs
SENya1990: I think it adds. But I would just split this method in two, each with its own switch, cause it's quite big

src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/CodeResolvingUtils.cs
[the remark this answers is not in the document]
SENya1990: @dropsonic ok

src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/CodeResolvingUtils.cs
SENya1990: I think it adds. But I would just split this method in two, each with its own switch, cause it's quite big

SENya1990: Actually in this case I'm not sure that it adds 2 levels of indentation. So the code style is not violated, but still two methods each with its own switch could look better

PR #87, opened 2018-09-03
ATR-221: edit regexp to match complex string format parameters

metallmonster:


src/Acuminator/Acuminator.Analyzers/StaticAnalysis/Localization/LocalizationMessageHelper.cs
```diff
@@ -12,7 +12,7 @@ namespace Acuminator.Analyzers.StaticAnalysis.Localization
 {
     internal class LocalizationMessageHelper
     {
-        private const string _formatRegexString = @"(?<Par>{(\w+|:+)})";
+        private const string _formatRegexString = @"(?<Par>{[\w:,-]+})";
```
dropsonic: Custom datetime format strings can contain any characters inside curly brackets:
https://docs.microsoft.com/en-us/dotnet/standard/base-types/custom-date-and-time-format-strings

dropsonic: Please take a look at the end of a table in the provided article.
It tells that a format string can contain "Any other character".
E.g. it can look like this: `{0:\Year\:YYYY, \Mon\t\h\: MM, Da\y: dd`}`, and for 2018-03-09 it outputs "Year: 2018, Month: 03, Day: 09".

metallmonster: @dropsonic thanks

PR #91, opened 2018-09-04
Feature/atr 90 changes in pxcache in event handlers

dropsonic:
Contains two features: ATR-90 and ATR-84.

### DIFF
--- a/src/Acuminator/Acuminator.Analyzers/StaticAnalysis/ChangesInPXCache/Walker.cs
+++ b/src/Acuminator/Acuminator.Analyzers/StaticAnalysis/ChangesInPXCache/Walker.cs
@@ -1,0 +1,68 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Acuminator.Utilities.Common;
+using Acuminator.Utilities.Roslyn;
+using Acuminator.Utilities.Roslyn.Semantic;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Acuminator.Analyzers.StaticAnalysis.ChangesInPXCache
+{
+	internal class Walker : NestedInvocationWalker
+	{
+		private static readonly ISet<string> MethodNames = new HashSet<string>(StringComparer.Ordinal)
+		{
+			"Insert" ,
+			"Update",
+			"Delete",
+		};
+
+		private SymbolAnalysisContext _context;
+		private readonly PXContext _pxContext;
+		private readonly DiagnosticDescriptor _diagnosticDescriptor;
+
+		public Walker(SymbolAnalysisContext context, PXContext pxContext, DiagnosticDescriptor diagnosticDescriptor)
+			: base(context.Compilation, context.CancellationToken)
+		{
+			pxContext.ThrowOnNull(nameof(pxContext));
+
+			_context = context;
+			_pxContext = pxContext;
+			_diagnosticDescriptor = diagnosticDescriptor;
+		}
+
+		public override void VisitInvocationExpression(InvocationExpressionSyntax node)
+		{
+			ThrowIfCancellationRequested();
+
+			var methodSymbol = GetSymbol<IMethodSymbol>(node);
+
+			if (methodSymbol != null && IsMethodForbidden(methodSymbol))
+			{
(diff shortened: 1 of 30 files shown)


src/Acuminator/Acuminator.Analyzers/StaticAnalysis/ChangesInPXCache/Walker.cs
```diff
@@ -0,0 +1,73 @@
...
+using Acuminator.Utilities.Roslyn.Semantic;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Acuminator.Analyzers.StaticAnalysis.ChangesInPXCache
+{
+	internal class Walker : NestedInvocationWalker
+	{
+		private static readonly ISet<string> MethodNames = new HashSet<string>(StringComparer.Ordinal)
```
SENya1990: @dropsonic what about "Clear"/"Remove" methods of PXCache?
Also shouldn't we consider the direct modifications of "Inserted"/"Updated"/"Deleted" collections?

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/SavingChanges/SaveOperationHelper.cs
```diff
@@ -0,0 +1,98 @@
...
+		private const string PressMethodName = "Press"; // PXAction.Press
+		private const string CachePersistMethodName = "Persist"; // PXCache.Persist
+		private const string CachePersistInserted = "PersistInserted"; // PXCache.PersistInserted
+		private const string CachePersistUpdated = "PersistUpdated"; // PXCache.PersistUpdated
+		private const string CachePersistDeleted = "PersistDeleted"; // PXCache.PersistDeleted
+
+		public static SaveOperationKind GetSaveOperationKind(IMethodSymbol symbol, InvocationExpressionSyntax syntaxNode, 
+			SemanticModel semanticModel, PXContext pxContext)
+		{
+			symbol.ThrowOnNull(nameof (symbol));
```
SENya1990: @dropsonic unusual indentation - do you have IDE setting which put space after `nameof`?

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/SavingChanges/SaveOperationHelper.cs
```diff
@@ -0,0 +1,98 @@
...
+			syntaxNode.ThrowOnNull(nameof (syntaxNode));
+			semanticModel.ThrowOnNull(nameof (semanticModel));
+			pxContext.ThrowOnNull(nameof (pxContext));
+
+			if (symbol.ContainingType != null)
+			{
+				// PXGraph.Actions.PressSave or PXSave.Press
+				if (String.Equals(symbol.Name, PressSaveMethodName, StringComparison.Ordinal)
```
SENya1990: @dropsonic are String.Equals necessary here? We are not working with UI strings and could rely on standard `==` which is ordinal by default and is shorter

SENya1990: @dropsonic also isn't a swith with pattern matching a little more readable?

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/ChangesInPXCache/Walker.cs
dropsonic: See the item in JIRA - I've asked Ruslan about that.
`Inserted` / `Updated` / `Deleted` are not collections, they are properties which return `IEnumerable` from the inner private collection which is inaccessible for direct modifications.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/SavingChanges/SaveOperationHelper.cs
dropsonic: It is a default setting from Resharper (for `nameof` and `typeof`).

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/SavingChanges/SaveOperationHelper.cs
dropsonic: @SENya1990 you're completely right, I'll fix it

PR #93, opened 2018-09-05
Feature/atr 79 master pxgraph createinstance in event handlers

dropsonic:


dropsonic:
Includes Bugfix/atr-224-master-recursive-analysis-reports-multiple-times

PR #95, opened 2018-09-08
Feature/event handler signature refactoring

dropsonic:


SENya1990:
@dropsonic Does this refactoring have  a detailed description somewhere?
We should either add a table with refactoring infos to Acumatica wiki or to the GitHub documentation. The process should be same as with diagnostics

src/Acuminator/Acuminator.Analyzers/Refactorings/EventHandlerSignature/EventHandlerSignatureRefactoring.cs
```diff
@@ -0,0 +1,160 @@
...
+using Microsoft.CodeAnalysis.CodeActions;
+using Microsoft.CodeAnalysis.CodeRefactorings;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Editing;
+using static Microsoft.CodeAnalysis.CSharp.SyntaxFactory;
+
+namespace Acuminator.Analyzers.Refactorings.EventHandlerSignature
+{
+	[ExportCodeRefactoringProvider(LanguageNames.CSharp), Shared]
```
SENya1990: @dropsonic Should the Name property of the attribute be set?

src/Acuminator/Acuminator.Analyzers/Refactorings/EventHandlerSignature/EventHandlerSignatureRefactoring.cs
```diff
@@ -0,0 +1,160 @@
...
+
+			if (methodNode != null)
+			{
+				IMethodSymbol methodSymbol = semanticModel.GetDeclaredSymbol(methodNode, context.CancellationToken);
+
+				if (IsSuitableForRefactoring(methodSymbol, pxContext))
+				{
+					var eventHandlerInfo = methodSymbol.GetEventHandlerInfo(pxContext);
+
+					if (eventHandlerInfo.eventSignatureType == EventHandlerSignatureType.Default
```
SENya1990: @dropsonic This condition is quite heavy to read even in IDE. you should put part of it in a variable like `key` or maybe put the whole check in a separate method

src/Acuminator/Acuminator.Analyzers/Refactorings/EventHandlerSignature/EventHandlerSignatureRefactoring.cs
```diff
@@ -0,0 +1,160 @@
...
+		{
+			return symbol?.ContainingType?.OriginalDefinition != null
+			       && symbol.ContainingType.OriginalDefinition.IsPXGraphOrExtension(pxContext)
+			       && symbol.Parameters.Length <= 2
+			       && !symbol.IsOverride
+				   && !symbol.GetAttributes().Any(
+					   attr => attr.AttributeClass != null && attr.AttributeClass.Equals(pxContext.AttributeTypes.PXOverrideAttribute));
+		}
+
+		private Task<Document> ChangeSignatureAsync(CodeRefactoringContext context,
```
SENya1990: @dropsonic you don't need to pass the whole `CodeRefactoringContext` to this method, you just need a Document.
Also there are a lot of parameters in this method. Can't their number be reduced like obtaining semantic model from Document once again? (we already removed caching of semantic model in other places).

src/Acuminator/Acuminator.Analyzers/Refactorings/EventHandlerSignature/EventHandlerSignatureRefactoring.cs
```diff
@@ -0,0 +1,160 @@
...
+			SyntaxNode root, SemanticModel semanticModel,
+			MethodDeclarationSyntax methodDeclaration, IMethodSymbol methodSymbol,
+			EventType eventType, INamedTypeSymbol genericArgsSymbol,
+			string dacName, string fieldName,
+			CancellationToken cancellationToken)
+		{
+			cancellationToken.ThrowIfCancellationRequested();
+
+			var newParameters = methodDeclaration.ParameterList.Parameters;
+			var cacheParameterSymbol = methodSymbol.Parameters[0];
```
SENya1990: @dropsonic, the `IsSuitableForRefactoring` method checks that the number of parameters in method <= 2
Are there some other checks which filter out methods with zero parameters? If not then such method will cause an exception.

src/Acuminator/Acuminator.Analyzers/Refactorings/EventHandlerSignature/EventHandlerSignatureRefactoring.cs
```diff
@@ -0,0 +1,160 @@
+﻿using System;
```
SENya1990: @dropsonic will the refactoring add corresponding usings if they are missing?

src/Acuminator/Acuminator.Analyzers/Refactorings/EventHandlerSignature/EventHandlerSignatureRefactoring.cs
```diff
@@ -0,0 +1,160 @@
...
+	{
+		private const string ArgsParameterName = "e";
+		private const string EventHandlerMethodName = "_";
+
+		public override async Task ComputeRefactoringsAsync(CodeRefactoringContext context)
+		{
+			context.CancellationToken.ThrowIfCancellationRequested();
+
+			var semanticModel = await context.Document.GetSemanticModelAsync(context.CancellationToken).ConfigureAwait(false);
+			var pxContext = new PXContext(semanticModel.Compilation);
```
SENya1990: @dropsonic no check of semantic model for null or cancellation.
Also no check (at least explicit) for the case when PX.Data is not present in the solution

src/Acuminator/Acuminator.Analyzers/Refactorings/EventHandlerSignature/EventHandlerSignatureRefactoring.cs
```diff
@@ -0,0 +1,160 @@
...
+				parameterName = argsParameter.Identifier;
+			}
+
+			newParameters = newParameters.Replace(argsParameter, 
+				CreateArgsParameter(genericArgsSymbol, parameterName, dacName, fieldName));
+
+			var newMethodDeclaration = 
+				ReplaceParameterUsages(methodDeclaration, cacheParameterSymbol,
+					MemberAccessExpression(SyntaxKind.SimpleMemberAccessExpression, 
+						IdentifierName(parameterName), IdentifierName("Cache")), semanticModel)
```
SENya1990: @dropsonic imho it's better to put all such literal strings into const fields.

src/Acuminator/Acuminator.Analyzers/Refactorings/EventHandlerSignature/EventHandlerSignatureRefactoring.cs
```diff
@@ -0,0 +1,160 @@
...
+
+			return Task.FromResult(context.Document.WithSyntaxRoot(root));
+		}
+
+		private ParameterSyntax CreateArgsParameter(INamedTypeSymbol genericArgsSymbol, SyntaxToken parameterName, 
+			string dacName, string fieldName)
+		{
+			TypeSyntax identifier;
+
+			if (String.IsNullOrEmpty(fieldName))
```
SENya1990: @dropsonic just in case - we have a little extension methods IsNullOrEmpty/IsNullOrWhiteSpace for strings which call corresponding String methods internally, so it could be written like `fieldName.IsNullOrEmpty()`

src/Acuminator/Acuminator.Analyzers/Refactorings/EventHandlerSignature/EventHandlerSignatureRefactoring.cs
```diff
@@ -0,0 +1,160 @@
...
+						IdentifierName(genericArgsSymbol.ContainingType.Name),
+						GenericName(
+								Identifier(genericArgsSymbol.Name))
+							.WithTypeArgumentList(
+								TypeArgumentList(
+									SingletonSeparatedList<TypeSyntax>(
+										identifier)))));
+		}
+
+		private (string dacName, string fieldName) ParseMethodName(string methodName, EventType eventType)
```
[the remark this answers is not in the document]
SENya1990: @dropsonic I didn't find much info about naming conventions about named tuples but those that I found suggest to use Pascal Case for returned tuples since it makes the elements of the returned object to look like properties which is consistent with .Net naming conventions:
https://www.infoq.com/articles/Patterns-Practices-CSharp-7

src/Acuminator/Acuminator.Utilities/Roslyn/Syntax/ParameterUsagesRewriter.cs
```diff
@@ -0,0 +1,49 @@
...
+		public override SyntaxNode VisitIdentifierName(IdentifierNameSyntax node)
+		{
+			var symbolInfo = _semanticModel.GetSymbolInfo(node);
+
+			if (symbolInfo.Symbol != null && symbolInfo.Symbol.Equals(_parameter))
+			{
+				var replacement = _replaceWith;
+				if (node.HasLeadingTrivia)
+					replacement = replacement.WithLeadingTrivia(node.GetLeadingTrivia());
+				if (node.HasTrailingTrivia)
```
SENya1990: @dropsonic no empty lines between control flow statements - we have rule for this in code style

dropsonic:
@SENya1990 Acumatica Wiki is not a right place for the documentation because it is not accessible from the outside. Descriptions for all diagnostics, refactorings, etc., will be added later, we have a task for that.

src/Acuminator/Acuminator.Analyzers/Refactorings/EventHandlerSignature/EventHandlerSignatureRefactoring.cs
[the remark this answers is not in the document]
dropsonic: @SENya1990 ok

src/Acuminator/Acuminator.Analyzers/Refactorings/EventHandlerSignature/EventHandlerSignatureRefactoring.cs
dropsonic: @SENya1990 nope, it is not required

src/Acuminator/Acuminator.Analyzers/Refactorings/EventHandlerSignature/EventHandlerSignatureRefactoring.cs
dropsonic: @SENya1990 it is possible, but it might affect performance

src/Acuminator/Acuminator.Analyzers/Refactorings/EventHandlerSignature/EventHandlerSignatureRefactoring.cs
dropsonic: @SENya1990 it is checked later using `GetEventHandlerInfo`

src/Acuminator/Acuminator.Analyzers/Refactorings/EventHandlerSignature/EventHandlerSignatureRefactoring.cs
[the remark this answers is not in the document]
dropsonic: @SENya1990 official blon on MSDN uses another naming convention:
https://blogs.msdn.microsoft.com/mazhou/2017/05/26/c-7-series-part-1-value-tuples/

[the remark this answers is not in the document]
dropsonic: but documentation uses PascalCase: https://docs.microsoft.com/en-us/dotnet/csharp/tuples

src/Acuminator/Acuminator.Analyzers/Refactorings/EventHandlerSignature/EventHandlerSignatureRefactoring.cs
dropsonic: @SENya1990 swapped conditions and added a comment to improve readability

src/Acuminator/Acuminator.Analyzers/Refactorings/EventHandlerSignature/EventHandlerSignatureRefactoring.cs
dropsonic: @SENya1990 I prefer not to use them because ReSharper shows "Possible NullReferenceException" warning in that case.

src/Acuminator/Acuminator.Analyzers/Refactorings/EventHandlerSignature/EventHandlerSignatureRefactoring.cs
SENya1990: @dropsonic it's better not to rely to much on internal implementation

src/Acuminator/Acuminator.Analyzers/Refactorings/EventHandlerSignature/EventHandlerSignatureRefactoring.cs
```diff
@@ -0,0 +1,175 @@
...
+using Acuminator.Utilities.Roslyn.Syntax;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CodeActions;
+using Microsoft.CodeAnalysis.CodeRefactorings;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Editing;
+using static Microsoft.CodeAnalysis.CSharp.SyntaxFactory;
+
+namespace Acuminator.Analyzers.Refactorings.EventHandlerSignature
```
SENya1990: @dropsonic It is not obvious that this refactoring convert old events into generic events. The event handler signature refactoring can mean a lot of things - maybe the name should be more concrete? Like ConvertToGenericEventHandler?

src/Acuminator/Acuminator.Analyzers/Refactorings/EventHandlerSignature/EventHandlerSignatureRefactoring.cs
SENya1990: @dropsonic we can add JetBrains Annotation attributes to them to fix the situation.

src/Acuminator/Acuminator.Analyzers/Refactorings/EventHandlerSignature/EventHandlerSignatureRefactoring.cs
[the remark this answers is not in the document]
SENya1990: @dropsonic the official from your first reference uses it too for the returned tuples. The returned tuples has Pascal Case names and in local variables they use Camel case. I think this depends on the context - where you use tuple and how. It's still not very clear documented(

src/Acuminator/Acuminator.Analyzers/Refactorings/EventHandlerSignature/EventHandlerSignatureRefactoring.cs
dropsonic: @SENya1990 I still didn't get it - what's wrong with cancellation?
Roslyn documentation doesn't state that `SemanticModel` can be null, and we've never faced such situation on practice.

src/Acuminator/Acuminator.Analyzers/Refactorings/EventHandlerSignature/EventHandlerSignatureRefactoring.cs
dropsonic: renamed

src/Acuminator/Acuminator.Analyzers/Refactorings/EventHandlerSignature/EventHandlerSignatureRefactoring.cs
dropsonic: @SENya1990 I couldn't find any annotation that tells ReSharper that this method performs a null checking.

PR #98, opened 2018-09-17
Bugfix/atr 217 master

SENya1990:


SENya1990:
Need changes to not show warnings about PXInt/PXDBInt on aggregators

PR #100, opened 2018-09-18
Feature/atr 78

metallmonster:


src/Acuminator/Acuminator.Analyzers/StaticAnalysis/LongOperationStart/StartLongOperationWalker.cs
```diff
@@ -0,0 +1,44 @@
...
+    public class StartLongOperationWalker : NestedInvocationWalker
+    {
+        private readonly PXContext _pxContext;
+        private readonly Action<Diagnostic> _reportDiagnostic;
+        private readonly DiagnosticDescriptor _descriptor;
+
+        public StartLongOperationWalker(SymbolAnalysisContext context, PXContext pxContext, DiagnosticDescriptor descriptor)
+            : base(context.Compilation, context.CancellationToken)
+        {
+            context.CancellationToken.ThrowIfCancellationRequested();
```
dropsonic: It is an overkill to check a cancellation token right in the constructor - it is checked immediately before the walker's creation.

metallmonster:
@dropsonic one more try

PR #104, opened 2018-09-19
Feature/atr 227 master release notes14

k-s-popova:
Please review the release notes document for 1.4.

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -1,6 +1,43 @@
 # Acuminator Release Notes
 This document provides information about fixes, enhancements, and key features that are available in Acuminator 1.3.
 
+## Acuminator 1.4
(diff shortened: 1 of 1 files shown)


docs/ReleaseNotes.md
```diff
@@ -1,20 +1,62 @@
 # Acuminator Release Notes
-This document provides information about fixes, enhancements, and key features that are available in Acuminator 1.3.
+This document provides information about fixes, enhancements, and key features that are available in Acuminator.
```
SENya1990: @k-s-popova @dropsonic I think everything is ok but we shouldn't release the new version without ATR-217 for which there is a pull request already. Too many false alerts for DACs otherwise. And if we add this item then the description for it should be added to the release notes

dropsonic:
Yes, everything is OK, but we have to include some other items into the upcoming release.

PR #109, opened 2018-09-20
Feature/atr 97 master row changes in row inserting

dropsonic:


src/Acuminator/Acuminator.Analyzers/StaticAnalysis/RowChangesInEventHandlers/DiagnosticWalker.cs
```diff
@@ -26,13 +26,16 @@ private class DiagnosticWalker : CSharpSyntaxWalker
 			private SymbolAnalysisContext _context;
 			private readonly SemanticModel _semanticModel;
 			private readonly PXContext _pxContext;
+			private readonly bool _reversed;
```
SENya1990: @dropsonic _reversed is not very clear name in code. Imho more meaningful name can be used. Even better option is a enum. Something like RowChangesAnalysisMode

dropsonic: @SENya1990 I agree, will try to come up with another name

PR #111, opened 2018-09-21
Feature/atr 95 dev database queries in row selected

dropsonic:


src/Acuminator/Acuminator.Utilities/CodeAnalysisSettings.cs
```diff
@@ -11,17 +11,51 @@ namespace Acuminator.Utilities
 	public class CodeAnalysisSettings
 	{
 		public static CodeAnalysisSettings Default => 
-			new CodeAnalysisSettings(recursiveAnalysisEnabled: true);
+			new CodeAnalysisSettings(
+				recursiveAnalysisEnabled: true, 
+				isvSpecificAnalyzersEnabled: false);
 
 		public CodeAnalysisSettings()
```
SENya1990: @dropsonic maybe the setting values shoud be the same as in the instance returned by the Default  property.
It feels inconsistent to have `new CodeAnalysisSettings()` with settings different from the  `CodeAnalysisSettings .Default`
If you plan to use it like VS Workspace Options then it could be like this `CodeAnalysisSettings .Default.WithRecursiveAnalysisEnabled()`. The default constructor feels redundant. Instead you could make the private constructor to be public

src/Acuminator/Acuminator.Utilities/CodeAnalysisSettings.cs
```diff
@@ -11,17 +11,51 @@ namespace Acuminator.Utilities
...
+		{
+			return WithIsvSpecificAnalyzersEnabledValue(true);
+		}
+
+		public CodeAnalysisSettings WithIsvSpecificAnalyzersDisabled()
+		{
+			return WithIsvSpecificAnalyzersEnabledValue(false);
+		}
+
+		protected virtual CodeAnalysisSettings WithRecursiveAnalysisEnabledValue(bool value)
```
SENya1990: @dropsonic are the othet methods really needed? You can just make this protected methods public

dropsonic: @SENya1990 yes, but it will decrease readability

src/Acuminator/Acuminator.Utilities/CodeAnalysisSettings.cs
dropsonic: @SENya1990 default constructor is used by the derived class which doesn't store those values - it gets them from the Options page

src/Acuminator/Acuminator.Utilities/CodeAnalysisSettings.cs
SENya1990: not so much  - compare `CodeAnalysisSettings.Default.WithIsvSpecificAnalyzersEnabledValue(false)` with
`CodeAnalysisSettings.Default.WithRecursiveAnalysisDisabled()`
I agree that second option is more readable but should we need to add three methods in the future for every new setting? The first option is quite clear too.

dropsonic: We rarely add new options, so it isn't a big deal.

src/Acuminator/Acuminator.Utilities/CodeAnalysisSettings.cs
SENya1990: But this is not the only one possible option to create default code analysis settings. If you want only derived classes to use the default constructor then it can be `protected`

src/Acuminator/Acuminator.Utilities/CodeAnalysisSettings.cs
SENya1990: Ok then

src/Acuminator/Acuminator.Utilities/CodeAnalysisSettings.cs
dropsonic: @SENya1990 default constructor is used in the unit tests

SENya1990: It is inconsistent to have the options set by the default constructor to be different from the ones set by the Default property. Imho they should be the same

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/DatabaseQueries/Walker.cs
```diff
@@ -0,0 +1,65 @@
...
+
+		private bool IsDatabaseCall(IMethodSymbol candidate)
+		{
+			var containingType = candidate.ContainingType?.OriginalDefinition;
+			return MethodPrefixes.Any(p => candidate.Name.StartsWith(p, StringComparison.Ordinal))
+				   && containingType != null
+				   && (containingType.IsBqlCommand(_pxContext)
+					   || containingType.InheritsFromOrEquals(_pxContext.PXViewType)
+					   || containingType.InheritsFromOrEquals(_pxContext.PXSelectorAttribute)
+					   || containingType.Equals(_pxContext.PXDatabase));
```
metallmonster: From my point of view this check is not correct logically and semantically.

1. I think here we should check pairs Type-Method instead of bulk check by type and bulk check by method after. Not all method names `MethodPrefixes` are applicable for `PXViewType` and for other types. For example, `PXView` does not have `Insert`, `Update`, `Delete` methods.
2. We can use `IMethodSymbol` for methods comparison instead of name comparison via string. For example, you compare `containingType` with `_pxContext.PXDatabase` not via strings, but via symbols
3. `PXDatabase.Insert` method should be checked too as it goes straight to DB.

dropsonic: @metallmonster this diagnostic reuses already existing code which I only moved in this branch. We can refactor it later if there is a need for it.

metallmonster: @dropsonic ok, but at least we need to create a bug to add `PXDatabase.Insert` method to this diagnostic

dropsonic: @metallmonster done

PR #117, opened 2018-09-25
Feature/atr 77

metallmonster:


src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/PXGraph/GraphSymbolUtils.cs
```diff
@@ -143,18 +143,161 @@ private static IEnumerable<INamedTypeSymbol> GetAllViewTypesFromPXGraphOrPXGraph
...
+            foreach (ISymbol member in graphOrExtension.GetMembers())
+            {
+                cancellation.ThrowIfCancellationRequested();
+
+                if (!(member is IMethodSymbol method))
+                {
+                    continue;
+                }
+
+                bool correctName = views.Any(v => v.Name.Equals(member.Name, StringComparison.OrdinalIgnoreCase));
```
SENya1990: @metallmonster what is the average views count in graph? Is it sensible to create hashset from the view names for this search?

src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/PXGraph/GraphSymbolUtils.cs
```diff
@@ -143,18 +143,161 @@ private static IEnumerable<INamedTypeSymbol> GetAllViewTypesFromPXGraphOrPXGraph
...
+                if (!correctName)
+                {
+                    continue;
+                }
+
+                bool correctReturnType = pxContext.SystemTypes.IEnumerable.Equals(method.ReturnType);
```
SENya1990: @metallmonster this doesn't consider the `IEnumerable<T>` return type. This must be handled.

src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/PXGraph/GraphSymbolUtils.cs
```diff
@@ -143,18 +143,161 @@ private static IEnumerable<INamedTypeSymbol> GetAllViewTypesFromPXGraphOrPXGraph
...
-		/// <summary>
-		/// Gets all declared view symbols and types from the graph and its base graphs,
-		/// if there is a graphs class hierarchy and <paramref name="includeViewsFromInheritanceChain"/> parameter is <c>true</c>.
-		/// </summary>
-		/// <param name="graph">The graph to act on.</param>
-		/// <param name="pxContext">Context.</param>
-		/// <param name="includeViewsFromInheritanceChain">(Optional) True to include, false to exclude the views from inheritance chain.</param>
-		/// <returns/>
-		public static ViewSymbolWithTypeCollection GetViewsWithSymbolsFromPXGraph(this ITypeSymbol graph, PXContext pxContext,
+        public static IEnumerable<(MethodDeclarationSyntax Node, IMethodSymbol Symbol)> GetViewDelegatesFromGraph(
```
SENya1990: @metallmonster this class became quite big. Maybe we should organaize it with regions? There are already regions for other methods? Also it's better to add xml descriptions to public helpers

src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/PXGraph/GraphSymbolUtils.cs
```diff
@@ -143,18 +143,161 @@ private static IEnumerable<INamedTypeSymbol> GetAllViewTypesFromPXGraphOrPXGraph
...
+            }
+
+            INamedTypeSymbol baseType = graphExtension.BaseType;
+            Dictionary<string, T> infoByView = new Dictionary<string, T>(StringComparer.OrdinalIgnoreCase);
+
+            for (int i = baseType.TypeArguments.Length - 1; i >= 0; i--)
+            {
+                ITypeSymbol argType = baseType.TypeArguments[i];
+
+                if (i == baseType.TypeArguments.Length - 1)
```
SENya1990: @metallmonster maybe the last type argument should be analyzed outside of the cycle and start iterations from the  `baseType.TypeArguments.Length - 2`. This will reduce the the complexity of the cycle as well as the number of comparison operations inside.

src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/ITypeSymbolExtensions.cs
```diff
@@ -330,7 +331,7 @@ private static string GetClrStyleTypeFullNameForNotNestedType(this ITypeSymbol n
             return initializers;
         }
 
-        public static (ConstructorDeclarationSyntax Node, IMethodSymbol Symbol) GetDeclaredStaticConstructor
+        public static StaticCtrInfo? GetDeclaredStaticConstructor
             (this INamedTypeSymbol typeSymbol, CancellationToken cancellation = default)
         {
```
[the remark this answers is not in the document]
dropsonic: @SENya1990 probably we should return them too because they are executed if you access a derived type

src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/PXGraph/GraphSymbolUtils.cs
```diff
@@ -143,18 +143,161 @@ private static IEnumerable<INamedTypeSymbol> GetAllViewTypesFromPXGraphOrPXGraph
...
+        }
+
+        private static IEnumerable<(MethodDeclarationSyntax Node, IMethodSymbol Symbol)> GetViewDelegatesFromGraphOrGraphExtension(
+            this ITypeSymbol graphOrExtension, IEnumerable<ISymbol> views, PXContext pxContext, CancellationToken cancellation)
+        {
+            foreach (ISymbol member in graphOrExtension.GetMembers())
+            {
+                cancellation.ThrowIfCancellationRequested();
+
+                if (!(member is IMethodSymbol method))
```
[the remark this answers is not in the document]
dropsonic: @metallmonster @SENya1990 in a recent pull request I've added a helper method `GetMethods` which does exactly what is needed here

src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/PXGraph/GraphSymbolUtils.cs
dropsonic: @metallmonster @SENya1990 I took a look on the platform code and it seems that a delegate can have only two signatures:
1. `IEnumerable` return type with any number of parameters, `ref` parameters are not allowed (`PXSelectDelegate`)
2. `void` return type with any number of parameters, but if any, they must be by `ref` (`PXPrepareDelegate`). This signature is used for parameters transformation.

Please note that a delegate can be either instance or static, public or non-public.

Generic `IEnumerable<T>` is not supported.

See `PXGraph.ProcessFields` method and  `PXView` constructor for more details.

src/Acuminator/Acuminator.Analyzers/Resources.resx
```diff
@@ -360,4 +360,7 @@
   <data name="PX1071Title" xml:space="preserve">
     <value>Actions cannot be executed within event handlers</value>
   </data>
+  <data name="PX1080Title" xml:space="preserve">
+    <value>A data view delegate should not start PXLongOperation</value>
```
dropsonic: Is this string verified?
Please see the corresponding string for event handlers - it uses different terms:

> A long-running operation cannot be started within an event handler

src/Acuminator/Acuminator.Tests/Tests/StaticAnalysis/LongOperationStart/LongOperationInDataViewDelegateTests.cs
```diff
@@ -0,0 +1,32 @@
...
+using Acuminator.Analyzers.StaticAnalysis.LongOperationStart;
+using Acuminator.Tests.Helpers;
+using Acuminator.Tests.Verification;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Xunit;
+using System.Threading.Tasks;
+
+namespace Acuminator.Tests.Tests.StaticAnalysis.LongOperationStart
+{
+    public class LongOperationInDataViewDelegateTests : DiagnosticVerifier
```
dropsonic: Please test some other cases:

1. Delegate with parameters
2. Static delegate
3. Delegate passed to `PXView`'s constructor (if this scenario is supported)

src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/ITypeSymbolExtensions.cs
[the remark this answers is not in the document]
dropsonic: @metallmonster I suggest using full name for the type `StaticConstructorInfo` because it is more readable, but it is not important

src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/PXGraph/DataViewInfo.cs
```diff
@@ -0,0 +1,16 @@
+﻿using Microsoft.CodeAnalysis;
+
+namespace Acuminator.Utilities.Roslyn.Semantic.PXGraph
+{
+    public readonly struct DataViewInfo
+    {
+        public ISymbol Symbol { get; }
```
dropsonic: It would be nice to add some XML comments here because it is not obvious what `Symbol` and `Type` mean

src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/PXGraph/GraphSymbolUtils.cs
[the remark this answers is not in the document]
SENya1990: @dropsonic yes, I've seen it. But it seems like he will always create an immutable array which is less perfomant. Maybe, because GetMembers already returns Immutable array that helper could return `IEnumerable<T>` since inside already underlies an ImmutableArray which can't be changed?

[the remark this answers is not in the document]
dropsonic: @SENya1990 all Roslyn methods return immutable collections, so I don't think we should consider it as "less performant" and worry about it at all.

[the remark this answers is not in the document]
SENya1990: ok

src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/ITypeSymbolExtensions.cs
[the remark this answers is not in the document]
metallmonster: Ok

src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/PXGraph/GraphSymbolUtils.cs
SENya1990: @dropsonic yes, you're right - checked it and it works only for IEnumerable.

src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/ITypeSymbolExtensions.cs
```diff
@@ -349,10 +350,10 @@ private static string GetClrStyleTypeFullNameForNotNestedType(this ITypeSymbol n
                 if (!(reference.GetSyntax(cancellation) is ConstructorDeclarationSyntax node))
```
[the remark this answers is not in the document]
metallmonster: Ok

src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/PXGraph/GraphSymbolUtils.cs
metallmonster: @SENya1990 I think there are a few dozens of views in a graph in general so it is not necessary to use a hashset here.

src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/PXGraph/GraphSymbolUtils.cs
metallmonster: @SENya1990 I agree that is a quite big. I don't like regions because if some part of the code is not related to the other part it is better to split them via files and classes, it my opinion. I'll create a separate file.
I agree about xml descriptions.

src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/PXGraph/GraphSymbolUtils.cs
SENya1990: @metallmonster a few dozens means around 30-40? I thought it was much less. Actuall Hashset is more perfomant in this case but the difference is not so big so decide yourself which is better.
According to this benchmark hashset becomes more performant  at about 20 items:
https://stackoverflow.com/questions/150750/hashset-vs-list-performance

src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/PXGraph/GraphSymbolUtils.cs
SENya1990: ok

src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/PXGraph/GraphSymbolUtils.cs
metallmonster: Ok

src/Acuminator/Acuminator.Analyzers/Resources.resx
metallmonster: No its not. I've requested Kseniya to review the string in Jira when I've created this pull-request.
Now the string is reviewed and I'll change it.

src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/PXGraph/DataViewInfo.cs
metallmonster: Ok

src/Acuminator/Acuminator.Tests/Tests/StaticAnalysis/LongOperationStart/LongOperationInDataViewDelegateTests.cs
metallmonster: 1. Ok
2. Ok
3. No, it is not supported

PR #134, opened 2018-10-03
Feature/atr 256 dev performance

dropsonic:
I've added some benchmarks that allow doing some rough comparison.
But they are not representative because Visual Studio has its own task scheduler and a lot of other background tasks that affect overall performance.

`Parallel.ForEach` seems to be 10% more effective in the benchmark, but my subjective tests on big solutions in Visual Studio show that it causes UI freezes.


SENya1990:
@dropsonic did you try to measure performance for PLinq?

src/Acuminator/Acuminator.Benchmark/RoslynHelper.cs
```diff
@@ -0,0 +1,35 @@
...
+namespace Acuminator.Benchmark
+{
+	public static class RoslynHelper
+	{
+		public static ImmutableArray<DiagnosticAnalyzer> GetAvailableAnalyzers()
+		{
+			return Assembly.GetAssembly(typeof(PXDiagnosticAnalyzer)).ExportedTypes
+				.Where(t => !t.IsAbstract && typeof(DiagnosticAnalyzer).IsAssignableFrom(t))
+				.Select(Activator.CreateInstance)
+				.Cast<DiagnosticAnalyzer>()
```
SENya1990: @dropsonic this way we analyze only the perfomance of the aggregators. Shouldn't we add more granulated benchmarks too?

dropsonic: We analyze performance for all analyzers, not for the aggregators.
Detailed telemetry is available in a separate project called "Acuminator.PerformanceReport".
But all these benchmarks are kinda useless because Visual Studio handles multithreading in its own very special way.

dropsonic:
@SENya1990 no, because there is no data flow that can be parallelized.

src/Acuminator/Acuminator.Benchmark/RoslynHelper.cs
SENya1990: @dropsonic but if I remeber correctly, only the aggregators derive from `PXDiagnosticAnalyzer`. 
Aggregated analyzers for PXGraph for example just implement `IPXGraphAnalyzer` interface. I haven't found the usage of this interface. Do you measure perfomance for such aggregated analyzers?

SENya1990:
@dropsonic the Parallel and PLinq have different implementation so it could be useful to try it too

src/Acuminator/Acuminator.Benchmark/RoslynHelper.cs
dropsonic: I got it. No, by now there is no such need.

dropsonic:
@SENya1990 the main issue is that we cannot measure the performance properly, so we cannot compare, e.g., Parallel vs PLinq.

src/Acuminator/Acuminator.Benchmark/RoslynHelper.cs
SENya1990: ok

SENya1990:
@dropsonic I agree, just was interested in comparison of benchmarks

dropsonic:
Closed since such optimizations have already been implemented and merged by @metallmonster.

PR #141, opened 2018-10-09
Feature/atr 244 legacy bql field

Derrekam:
Implemented IBqlField -> BqlTypedField and Constant<T> -> BqlTypedConstant diagnostics and fixes.
Using them in batch mode will help to migrate from BQL to FBQL


dropsonic:
@Derrekam are you sure that it should be a diagnostic with the Info severity? Code Refactoring seems like a more proper type to such operations.

PR #150, opened 2018-10-23
ATR-237: fixed the code fix name, refactored the attribute, added can…

SENya1990:
…cellation support

### DIFF
--- a/src/Acuminator/Acuminator.Analyzers/StaticAnalysis/MissingTypeListAttribute/MissingTypeListAttributeFix.cs
+++ b/src/Acuminator/Acuminator.Analyzers/StaticAnalysis/MissingTypeListAttribute/MissingTypeListAttributeFix.cs
@@ -14,52 +14,51 @@ using Microsoft.CodeAnalysis.CSharp.Syntax;
 
 namespace Acuminator.Analyzers.StaticAnalysis.MissingTypeListAttribute
 {
-    [ExportCodeFixProvider(LanguageNames.CSharp), Shared]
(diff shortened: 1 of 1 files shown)


src/Acuminator/Acuminator.Analyzers/StaticAnalysis/MissingTypeListAttribute/MissingTypeListAttributeFix.cs
```diff
@@ -24,48 +24,61 @@ public class MissingTypeListAttributeFix : CodeFixProvider
 
         public override async Task RegisterCodeFixesAsync(CodeFixContext context)
         {
-            var root = await context.Document.GetSyntaxRootAsync().ConfigureAwait(false);
-            var node = (PropertyDeclarationSyntax)root.FindNode(context.Span);
+			context.CancellationToken.ThrowIfCancellationRequested();
```
dropsonic: tabs/spaces mix

SENya1990: fixed

dropsonic:
@SENya1990 same thing as in the previous pull request - GitHub shows that the spaces were replaced with tabs

SENya1990:
@dropsonic isn't it our code style formatting setting to use Keep tabs option and Smart Indent?

dropsonic:
@SENya1990 I think so but in this case, the old formatting has been changed (in places where there were no changes at all), so it makes it extremely difficult to compare

PR #156, opened 2018-10-30
Bugfix/atr 275 dev refactoring fails

SENya1990:


dropsonic:
@SENya1990 I support there is a similar check for the BQL formatter command, is it?

SENya1990:
@dropsonic I don't know) Maybe not, I'll check

SENya1990:
@dropsonic  No there is no check. Actually the formatter doesn't use PXContext but it's own BqlContext which is a duplication and probably should be refactored in the future

PR #157, opened 2018-10-30
Bugfix/atr 278 dev coloriser fails

SENya1990:


dropsonic:
@SENya1990 as far as I remember, Roslyn 1.3.1 contains some new features that have been used in the analyzers (e.g. code block actions). Does they still work after the downgrade?

SENya1990:
@dropsonic as far as I remember they are disabled. All active unit tests passed but I didn't check them specifically. Anyway it's not like we have a choice since we have to support VS2015 and the downgrade of the System.Collections.Immutable package caused the downgrade of CodeAnalysis packages as well. 
I also don't know what will happen if you manually set the CodeAnalysis packages version to 1.3.1 bypassing Nuget but I suspect it won't work since CodeAnalysis declares dependancy on the newer version of the  System.Collections.Immutable.
There are still "hacky" options though like doing something during assembly resolving event, maybe you can add  System.Collections.Immutable to the package as a Content or Resource but imho it's better to avoid such things

PR #160, opened 2018-11-06
Feature/atr 252

metallmonster:


src/Acuminator/Acuminator.Analyzers/StaticAnalysis/ThrowingExceptions/WalkerForGraphAnalyzer.cs
```diff
@@ -0,0 +1,35 @@
...
+            descriptor.ThrowOnNull(nameof(descriptor));
+
+            _descriptor = descriptor;
+        }
+
+        public override void VisitThrowStatement(ThrowStatementSyntax node)
+        {
+            ThrowIfCancellationRequested();
+
+            if (IsPXSetupNotEnteredException(node))
```
SENya1990: @metallmonster please consider enhancing of `IsPXSetupNotEnteredException` method.
Currently it starts like this:
```
if (node == null || !(node.Expression is ObjectCreationExpressionSyntax objCreationSyntax) ||
                objCreationSyntax.Type == null)
            {
                return false;
            }
```
This leaves only expressions like:
`throw new MyExceptionType();`
It would be good to support also variables stored in exceptions:
```
var exception = new MyExceptionType();
PXTrace.WriteError(exception);
throw exception ;
```

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/ThrowingExceptions/WalkerBase.cs
```diff
@@ -40,5 +39,22 @@ protected bool IsPXSetupNotEnteredException(ThrowStatementSyntax node)
             var isSetupNotEntered = exceptionType.InheritsFromOrEquals(_pxContext.Exceptions.PXSetupNotEnteredException);
             return isSetupNotEntered;
         }
+
+        private ITypeSymbol GetExceptionType(ExpressionSyntax node)
```
SENya1990: @metallmonster I think we should refactor out and share a part of this method's logic with a switch and place it in Semantic Utils. It could be an extension method, like `GetSymbolType` (with a possibility of further extension)

metallmonster: @SENya1990 I was able to achieve the same logic without `GetExceptionType` method: I used `semanticModel.GetTypeInfo(node.Expression)`, it works fine, so I will delete `GetExceptionType` method

metallmonster:
@SENya1990 one more try

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/ThrowingExceptions/WalkerBase.cs
SENya1990: ok

PR #161, opened 2018-11-06
Bugfix/atr 239 master

SENya1990:


src/Acuminator/Acuminator.Analyzers/StaticAnalysis/ViewDeclarationOrder/ViewDeclarationOrderAnalyzer.cs
```diff
@@ -1,77 +1,124 @@
...
-    {
-        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics =>
-            ImmutableArray.Create(Descriptors.PX1004_ViewDeclarationOrder, Descriptors.PX1006_ViewDeclarationOrder);
+	/// <summary>
+	/// An analyzer for the order of view declaration in graph/graph extension.
+	/// This diagnostic works only for simple class hierarchy where the depth of the inheritance is equal to 2: object -> DAC -> DerivedDAC.
+	/// </summary>
+	public class ViewDeclarationOrderAnalyzer : IPXGraphAnalyzer
+	{
+		private const string InitCacheMappingMethodName = "InitCacheMapping";
```
metallmonster: It is better to use nameof() instead of string constant

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/ViewDeclarationOrder/ViewDeclarationOrderAnalyzer.cs
```diff
@@ -1,77 +1,124 @@
...
 				return;
 
-			//backward pass
-			CheckGraphViews(context, graphLocation, graphViews.Reverse(), Descriptors.PX1006_ViewDeclarationOrder);
+			foreach (IGrouping<ITypeSymbol, DataViewInfo> dacViews in viewsGroupedByDAC)
+			{
+				AnalyzeDacViewsForNumberOfCaches(graphSemanticModel, symbolContext, dacViews, viewsGroupedByDAC);
+			}
+
+			symbolContext.CancellationToken.ThrowIfCancellationRequested();
```
metallmonster: What is the reason to check cancellation token at the end of the method? It will decrease performance.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/ViewDeclarationOrder/ViewDeclarationOrderAnalyzer.cs
```diff
@@ -1,77 +1,124 @@
...
-											DiagnosticDescriptor diagnosticDescriptor)
+		/// <summary>
+		/// Starting from the Acumatica 2018R2 version a new method is used to initialize caches with explicit ordering of caches.
+		/// </summary>
+		/// <returns/>
+		private static bool IsNewMethodUsedToInitCaches(PXContext pxContext)
 		{
-			var visitedViews = new HashSet<ITypeSymbol>();
+			IMethodSymbol initCachesNewMethod = pxContext.PXGraph.Type
+																 .GetMembers(InitCacheMappingMethodName)
```
metallmonster: In such cases we place symbols to the appropriate utility classes. In this case, you can use `Acuminator.Utilities.Roslyn.Semantic.Symbols.PXGraphSymbols` class to store the method symbol.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/ViewDeclarationOrder/ViewDeclarationOrderAnalyzer.cs
SENya1990: This was done with the consideration that we will move from PX.Data to string constants...  somewhere in the future. In that case nameof won't work.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/ViewDeclarationOrder/ViewDeclarationOrderAnalyzer.cs
SENya1990: ok

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/ViewDeclarationOrder/ViewDeclarationOrderAnalyzer.cs
SENya1990: I wasn't sure - because this symbol  probably will be used by only this diagnostic. But I agree - it's cleaner this way. Will fix

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/ViewDeclarationOrder/ViewDeclarationOrderAnalyzer.cs
metallmonster: As far as I know, currently there is no plans to do it.

SENya1990:
Will reopen after introducing ShouldAnalyze to diagnostics

PR #165, opened 2018-11-12
Bugfix/atr 289 master fixed recursion

SENya1990:


SENya1990:
A special branch created from master to add recursion bugfix to 1.5

PR #166, opened 2018-11-13
Bugfix/atr 239 master

SENya1990:
Nothing new was actually added to this branch since previous pull request


src/Acuminator/Acuminator.Tests/Tests/StaticAnalysis/ViewDeclarationOrder/ViewDeclarationOrderTests.cs
```diff
@@ -1,23 +1,60 @@
...
 
 namespace Acuminator.Tests.Tests.StaticAnalysis.ViewDeclarationOrder
 {
     public class ViewDeclarationOrderTests : CodeFixVerifier
     {
-	    protected override DiagnosticAnalyzer GetCSharpDiagnosticAnalyzer() => new ViewDeclarationOrderAnalyzer();
+	    protected override DiagnosticAnalyzer GetCSharpDiagnosticAnalyzer() =>  
+			new PXGraphAnalyzer(CodeAnalysisSettings.Default
+													.WithRecursiveAnalysisEnabled()
+													.WithIsvSpecificAnalyzersEnabled(),
```
dropsonic: Is PX1004/PX1006 analyzer affected by ISV-Only options?

src/Acuminator/Acuminator.Tests/packages.config
```diff
@@ -22,18 +22,18 @@
...
-  <package id="System.Reflection" version="4.3.0" targetFramework="net461" />
+  <package id="System.Linq" version="4.3.0" targetFramework="net461" requireReinstallation="true" />
+  <package id="System.Linq.Expressions" version="4.3.0" targetFramework="net461" requireReinstallation="true" />
+  <package id="System.Reflection" version="4.3.0" targetFramework="net461" requireReinstallation="true" />
   <package id="System.Reflection.Metadata" version="1.2.0" targetFramework="net461" />
   <package id="System.Reflection.Primitives" version="4.3.0" targetFramework="net461" />
   <package id="System.Resources.ResourceManager" version="4.3.0" targetFramework="net461" />
-  <package id="System.Runtime" version="4.3.0" targetFramework="net461" />
-  <package id="System.Runtime.Extensions" version="4.3.0" targetFramework="net461" />
+  <package id="System.Runtime" version="4.3.0" targetFramework="net461" requireReinstallation="true" />
```
dropsonic: Why do we need `requireReinstallation="true"` here for all these packages?

src/Acuminator/Acuminator.Tests/Tests/StaticAnalysis/ViewDeclarationOrder/ViewDeclarationOrderTests.cs
SENya1990: No, because it was one of the first diagnostic from the hackathon. We can add this if we want, but it should be documented

src/Acuminator/Acuminator.Tests/packages.config
SENya1990: Frankly speaking, I have no idea because this config file were auto generated when I tried to emove System.ValueTuples package

src/Acuminator/Acuminator.Tests/Tests/StaticAnalysis/ViewDeclarationOrder/ViewDeclarationOrderTests.cs
dropsonic: Then it should not be here. The default behavior is fine, and the test doesn't require specifying this explicitly.

SENya1990: ok

PR #167, opened 2018-11-13
Feature/atr 282 should analyze

SENya1990:
Merge after ATR-239


src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/PXGraph/Infos/DataViewInfo.cs
```diff
@@ -20,6 +20,8 @@ public class DataViewInfo : GraphNodeSymbolItem<ISymbol>
 		/// The type of the data view symbol
 		/// </summary>
 		public INamedTypeSymbol Type { get; }
+		
+		public ITypeSymbol ViewDAC { get; }
```
metallmonster: What is the purpose of this property?

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/CallingBaseDataViewDelegate/CallingBaseDataViewDelegateFromOverrideDelegateAnalyzer.cs
```diff
@@ -18,7 +18,9 @@ public class CallingBaseDataViewDelegateFromOverrideDelegateAnalyzer : IPXGraphA
         public ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics =>
             ImmutableArray.Create(Descriptors.PX1087_CausingStackOverflowExceptionInBaseViewDelegateInvocation);
 
-        public void Analyze(SymbolAnalysisContext context, PXContext pxContext, CodeAnalysisSettings settings, PXGraphSemanticModel pxGraph)
+		public virtual bool ShouldAnalyze(PXContext pxContext, CodeAnalysisSettings settings) => true;
```
metallmonster: I think it is better to add a base class with this method returning true instead of copy-pasting it in most of the analyzers

src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/PXGraph/Infos/DataViewInfo.cs
SENya1990: It stores the DAC of the view. For example:
```
PXSelect<TDAC> -view
TDAC - ViewDAC
```
 It's quite convenient to have this info in DTO instead of retrieving it from context

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/CallingBaseDataViewDelegate/CallingBaseDataViewDelegateFromOverrideDelegateAnalyzer.cs
SENya1990: Yes, I agree. I'll think about a best way to add base aggregated analyzer class to system

SENya1990:
@metallmonster one more try

PR #175, opened 2018-11-19
Feature/atr 287 master more release notes15

k-s-popova:
Please review addition to release notes and new diagnostic descriptions

### DIFF
--- a/docs/diagnostics/PX1055.md
+++ b/docs/diagnostics/PX1055.md
@@ -1,0 +1,68 @@
+# PX1055
+This document describes the PX1055 diagnostic.
+
+## Summary
+
+| Code   | Short Description                          | Type  | Code Fix  | 
+| ------ | ------------------------------------------ | ----- | --------- | 
+| PX1055 | An invalid primary key of the DAC is used. | Error | Available | 
+
+## Diagnostic Description
+The key fields of a DAC can be defined as follows:
+
+ - A field bound to an identity column (that is, a field that has the `PXDBIdentity` or `PXDBLongIdentity` attribute) can be the only key field in the DAC.
+ - A DAC can have multiple key fields if none of these key fields is bound to an identity column.
+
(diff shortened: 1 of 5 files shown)


docs/diagnostics/PX1088.md
```diff
@@ -0,0 +1,44 @@
...
+            Where<SOShipmentExtension.aSNSent, Equal<False>,
+                Or<SOShipmentExtension.aSNSent, IsNull>>,
+            And2<
+                Where<SOShipmentExtension.readyToSend, Equal<True>>,
+                And<SOShipmentExtension.shipmentTypePkey, IsNotNull>>>>
+        Shipments;
+    public PXSetup<CustomerSettings> CustomerSettings;
+    public PXSelect<SOShipment> AllShipments;
+
+    public void SomeMethod()
```
dropsonic: It is not clear that `SomeMethod()` is a processing delegate.
I think that the constructor with `Shipments.SetProcessDelegate(SomeMethod)` should be added.

docs/diagnostics/PX1055.md
```diff
@@ -0,0 +1,68 @@
...
+
+| Code   | Short Description                          | Type  | Code Fix  | 
+| ------ | ------------------------------------------ | ----- | --------- | 
+| PX1055 | An invalid primary key of the DAC is used. | Error | Available | 
+
+## Diagnostic Description
+The key fields of a DAC can be defined based on the following guidelines:
+
+ - A field bound to an identity column (that is, a field that has the `PXDBIdentity` or `PXDBLongIdentity` attribute) can be the only key field in the DAC.
+ - A DAC can have multiple key fields if none of these key fields is bound to an identity column.
```
dropsonic: It is not clear that a DAC can have one non-identity key field (very common case).
The purpose of this diagnostic is to avoid mixing identity and non-identity fields in the primary key.

PR #178, opened 2018-12-19
Feature/atr 244 legacy bql field constant

Derrekam:


Derrekam:
Looks like code refactoring doesn't have "Fix all" option
https://github.com/dotnet/roslyn/issues/32461
Let's keep current items as diagnostic+fix with info level as they are now?

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/LegacyBqlField/LegacyBqlFieldAnalyzer.cs
```diff
@@ -0,0 +1,78 @@
...
+		protected override bool ShouldAnalyze(PXContext pxContext) => pxContext.IsAcumatica2019R1;
+
+		internal override void AnalyzeCompilation(CompilationStartAnalysisContext compilationStartContext, PXContext pxContext)
+		{
+			compilationStartContext.RegisterSymbolAction(
+				c => Analyze(c, pxContext),
+				SymbolKind.NamedType);
+		}
+
+		private async void Analyze(SymbolAnalysisContext context, PXContext pxContext)
```
SENya1990: async is not needed here:
```suggestion
		private void Analyze(SymbolAnalysisContext context, PXContext pxContext)
```

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/LegacyBqlField/LegacyBqlFieldAnalyzer.cs
```diff
@@ -0,0 +1,78 @@
...
+		{
+			if (context.Symbol is INamedTypeSymbol dacFieldType)
+			{
+				if (!IsDacFieldType(dacFieldType, pxContext) || AlreadyStronglyTyped(dacFieldType, pxContext))
```
SENya1990: @Derrekam this diagnostic should run for DACs and analyze all fields of a DAC/ DAC extension. It will be far more perfomant than analyze each fialed of the same DAC separately and perform all required checks. Moreover, if you analyze whole DAC you can remove some of the checks in `IsDacFieldType` method like:
```
dacFieldType.ContainingType != null &&
dacFieldType.ContainingType.IsDacOrExtension(pxContext);
```

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/LegacyBqlField/LegacyBqlFieldAnalyzer.cs
```diff
@@ -0,0 +1,78 @@
...
+						context.ReportDiagnostic(Diagnostic.Create(Descriptors.PX1060_LegacyBqlField, location, properties.ToImmutable(), dacFieldType.Name));
+					}
+				}
+			}
+		}
+
+		private static bool IsDacFieldType(ITypeSymbol dacFieldType, PXContext pxContext)
+			=> dacFieldType.TypeKind == TypeKind.Class &&
+				dacFieldType.IsDacField() &&
+				dacFieldType.BaseType.SpecialType == SpecialType.System_Object &&
```
SENya1990: @Derrekam  why do you check if the base type is object?

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/LegacyBqlField/LegacyBqlFieldAnalyzer.cs
```diff
@@ -0,0 +1,78 @@
...
+				dacFieldType.ContainingType != null &&
+				dacFieldType.ContainingType.IsDacOrExtension(pxContext);
+
+		internal static bool AlreadyStronglyTyped(INamedTypeSymbol dacFieldType, PXContext pxContext)
+			=> dacFieldType.AllInterfaces.Any(t =>
+				t.IsGenericType
+				&& t.ConstructUnboundGenericType().Name == pxContext.IImplementType.Name
```
SENya1990: @Derrekam why didn't you used `OriginalDefinition` instead of `ConstructUnboundGenericType()`

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/LegacyBqlField/LegacyBqlFieldFix.cs
```diff
@@ -0,0 +1,87 @@
...
+											SimpleBaseType(CreateBaseType(typeName, classNode.Identifier.Text))
+										}))))),
+					title),
+				context.Diagnostics);
+		}
+
+		private TypeSyntax CreateBaseType(string typeName, string dacFieldName)
+		{
+			if (!PropertyTypeToFieldType.ContainsKey(typeName))
+				throw new NotSupportedException();
```
SENya1990: @dropsonic, @metallmonster I don't remember will the VS catch the code fix not implemented exception?

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/LegacyBqlConstant/LegacyBqlConstantFix.cs
```diff
@@ -0,0 +1,71 @@
...
+										{
+											SimpleBaseType(CreateBaseType(typeName, classNode.Identifier.Text))
+										}))))),
+					title),
+				context.Diagnostics);
+		}
+
+		private TypeSyntax CreateBaseType(string typeName, string dacFieldName)
+		{
+			if (!LegacyBqlFieldFix.PropertyTypeToFieldType.ContainsKey(typeName))
```
SENya1990: @dropsonic, @metallmonster I don't remember will the VS catch the code fix not implemented exception?

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/LegacyBqlField/LegacyBqlFieldFix.cs
```diff
@@ -0,0 +1,87 @@
...
+
+			context.CancellationToken.ThrowIfCancellationRequested();
+
+			string title = nameof(Resources.PX1060Fix).GetLocalized().ToString();
+
+			context.RegisterCodeFix(
+				CodeAction.Create(
+					title,
+					c => Task.FromResult(
+						context.Document.WithSyntaxRoot(
```
SENya1990: @Derrekam  if you think that this code fix will be used in multiple place at once via Fix All in Solution/Project/Document then you should use Roslyn Simplifier to simplify names of the fields. It will be quite unconvenient to delete PX.Data.BQL namespace prefixes in the whole solution. I think ideally the fix should add the `using PX.Data.BQL` to the list of usings if there isn't one already.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/LegacyBqlField/LegacyBqlFieldFix.cs
```diff
@@ -0,0 +1,87 @@
...
+
+		private TypeSyntax CreateBaseType(string typeName, string dacFieldName)
+		{
+			if (!PropertyTypeToFieldType.ContainsKey(typeName))
+				throw new NotSupportedException();
+
+			return IdentifierName($"PX.Data.BQL.Bql{PropertyTypeToFieldType[typeName]}.Field<{dacFieldName}>");
+		}
+
+		public static readonly IReadOnlyDictionary<string, string> PropertyTypeToFieldType = new Dictionary<string, string>
```
SENya1990: @Derrekam this register currently doesn't support string arrays, UInt16 arrays and float. We have such attributes and DAC properties - see FieldTypeAttributesRegister.cs class.

src/Acuminator/Acuminator.Tests/Tests/StaticAnalysis/LegacyBqlField/Sources/LegacyBqlFieldBad.cs
```diff
@@ -0,0 +1,16 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using PX.Data;
+
+namespace PX.Analyzers.Test.Sources
+{
+	public class BadDac : IBqlTable
+	{
+		public abstract class legacyField : IBqlField { }
```
SENya1990: @Derrekam please add more fields to the test. It will be good to test field of byte[] array type, field of primitive type like int, field of primitive type which is written with .Net name like Int32, maybe some exotic (uint16[]) or not supported fields.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/LegacyBqlConstant/LegacyBqlConstantFix.cs
```diff
@@ -0,0 +1,71 @@
...
+			}
+
+			context.CancellationToken.ThrowIfCancellationRequested();
+
+			string title = nameof(Resources.PX1061Fix).GetLocalized().ToString();
+
+			context.RegisterCodeFix(
+				CodeAction.Create(
+					title,
+					c => Task.FromResult(
```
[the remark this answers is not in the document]
dropsonic: It is better to make it async.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/LegacyBqlField/LegacyBqlFieldFix.cs
dropsonic: It will lead to an unhandled exception. If some case is not supported, the code fix should not be registered at all.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/LegacyBqlField/LegacyBqlFieldAnalyzer.cs
Derrekam: to be sure that legacy field doesn't inherit from any class, because fix will replace IBqlField with proper Bql[Type].Field<TSelf> class, and we don't want to break existing code that uses fields in unusual way

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/LegacyBqlField/LegacyBqlFieldAnalyzer.cs
Derrekam: Would it be more efficient to use ```OriginalDefinition```?

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/LegacyBqlField/LegacyBqlFieldFix.cs
Derrekam: there is no representation for such type of fields in FBQL, therefore there is no need to provide such support in acuminator

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/LegacyBqlField/LegacyBqlFieldAnalyzer.cs
Derrekam: And how many fixes should "mass" diagnostic provide? One for a whole DAC, or many for every single field?

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/LegacyBqlConstant/LegacyBqlConstantAnalyzer.cs
```diff
@@ -0,0 +1,70 @@
...
+				{
+					var properties = ImmutableDictionary.CreateBuilder<string, string>();
+					properties.Add(CorrespondingType, constantType);
+
+					context.ReportDiagnostic(Diagnostic.Create(Descriptors.PX1061_LegacyBqlConstant, location, properties.ToImmutable(), constant.Name)); 
+				}
+			}
+		}
+
+		private static bool IsConstant(ITypeSymbol constantDef, PXContext pxContext, out string constantType)
```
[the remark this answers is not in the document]
Derrekam: I believe that this helper is very specific to the current diagnostic, and should not be added to shared utils.
In my opinion, both approaches are not as convenient to use as the current one

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/LegacyBqlConstant/LegacyBqlConstantFix.cs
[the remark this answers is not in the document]
Derrekam: so how should I change it exactly? I peeked this approach in some other diagnostic and copied it as-is

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/LegacyBqlField/LegacyBqlFieldFix.cs
Derrekam: Is there an example in existing diagnostics that I can use?

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/LegacyBqlConstant/LegacyBqlConstantFix.cs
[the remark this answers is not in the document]
SENya1990: I would suggest to use it inside Task.Run.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/LegacyBqlConstant/LegacyBqlConstantAnalyzer.cs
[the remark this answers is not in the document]
SENya1990: Ok, it's a matter of taste about signature, but can you explain why this helper is specific to your diagnostic?
It determines not every BQL constants but only some special ones? Should we update our helper?

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/LegacyBqlField/LegacyBqlFieldAnalyzer.cs
SENya1990: @Derrekam  I think, it depends on your vision of your feature. The more detailed approach is to register code fix for every DAC field, but on the other side it will be strange to make fix for one DAC field and don't make it for another. 
Still it is good to show warning not on the DAC name but on the specific field.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/LegacyBqlField/LegacyBqlFieldFix.cs
SENya1990: @Derrekam unfortunately there is only a little example in Acuminator in `SyntaxNodeGenerationUtils.GetAttributeList method. You can find more info here:
https://docs.microsoft.com/en-us/dotnet/csharp/roslyn-sdk/tutorials/how-to-write-csharp-analyzer-code-fix (the end of the manual)
https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.simplification.simplifier?view=roslyn-dotnet

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/LegacyBqlField/LegacyBqlFieldFix.cs
SENya1990: Ok, but you need to make sure that the exception won't be thrown for such types. I suggest to add such fields to DAC in source files for unit tests

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/LegacyBqlField/LegacyBqlFieldAnalyzer.cs
SENya1990: Actually, yes. According to Roslyn reference source for C# `ConstructUnboundGenericType()` internally looks like this
```
 /// <summary>
 /// Returns an unbound generic type of this named type.
 /// </summary>
 public NamedTypeSymbol ConstructUnboundGenericType()
  {
       return OriginalDefinition.AsUnboundGenericType();
  }
```
The `AsUnboundGenericType` internals:
```
static public NamedTypeSymbol AsUnboundGenericType(this NamedTypeSymbol type)
{
	if (!type.IsGenericType)
	{
		// This exception is part of the public contract of NamedTypeSymbol.ConstructUnboundGenericType
		throw new InvalidOperationException();
	}

	var original = type.OriginalDefinition;
	int n = original.Arity;
	NamedTypeSymbol originalContainingType = original.ContainingType;

	var constructedFrom = ((object)originalContainingType == null) ?
		original :
		original.AsMember(originalContainingType.IsGenericType ? originalContainingType.AsUnboundGenericType() : originalContainingType);
	if (n == 0)
	{
		return constructedFrom;
	}

	var typeArguments = UnboundArgumentErrorTypeSymbol.CreateTypeArguments(
		constructedFrom.TypeParameters,
		n,
		new CSDiagnosticInfo(ErrorCode.ERR_UnexpectedUnboundGenericName));
	return constructedFrom.Construct(typeArguments, unbound: true);
}
```
As you can see there is a lot of additional actions for `ConstructUnboundGenericType`. We didn't encounter any problems with 'OriginalDefinition' so I think you can safely use it.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/LegacyBqlField/LegacyBqlFieldAnalyzer.cs
SENya1990: @Derrekam  I have sent my thoughts about possible implementations to you in Slack

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/LegacyBqlConstant/LegacyBqlConstantAnalyzer.cs
```diff
@@ -0,0 +1,70 @@
...
+
+				context.CancellationToken.ThrowIfCancellationRequested();
+
+				Location location = constant.Locations.FirstOrDefault();
+				if (location != null)
+				{
+					var properties = ImmutableDictionary.CreateBuilder<string, string>();
+					properties.Add(CorrespondingType, constantType);
+
+					context.ReportDiagnostic(Diagnostic.Create(Descriptors.PX1061_LegacyBqlConstant, location, properties.ToImmutable(), constant.Name));
```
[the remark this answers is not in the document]
Derrekam: done

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/LegacyBqlConstant/LegacyBqlConstantAnalyzer.cs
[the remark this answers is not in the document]
Derrekam: It determines only old-fashioned constant and also provides info for creating a corresponding FBQL constant

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/LegacyBqlConstant/LegacyBqlConstantFix.cs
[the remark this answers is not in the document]
Derrekam: done

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/LegacyBqlField/LegacyBqlFieldFix.cs
Derrekam: replaced with return null

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/LegacyBqlConstant/LegacyBqlConstantFix.cs
Derrekam: replaced with return null

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/LegacyBqlConstant/LegacyBqlConstantFix.cs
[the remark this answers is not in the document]
dropsonic: `Task.Run` in this case is not recommended, as stated in our Best Practices.
You can find an example here: https://msdn.microsoft.com/en-us/magazine/dn904670.aspx?f=255&MSPPError=-2147217396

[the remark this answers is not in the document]
SENya1990: @dropsonic then what should be used for code fix which doesn't contain any `await` operations?
Is there a difference between returning `Task.Run` and `await/async`? Maybe the synchronious pattern is ok?

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/LegacyBqlField/LegacyBqlFieldAnalyzer.cs
```diff
@@ -25,23 +25,29 @@ internal override void AnalyzeCompilation(CompilationStartAnalysisContext compil
 
 		private async void Analyze(SymbolAnalysisContext context, PXContext pxContext)
 		{
-			if (context.Symbol is INamedTypeSymbol dacFieldType)
+			if (context.Symbol is INamedTypeSymbol dacOrDacExt)
 			{
-				if (!IsDacFieldType(dacFieldType, pxContext) || AlreadyStronglyTyped(dacFieldType, pxContext))
+				if (dacOrDacExt == null || (!dacOrDacExt.IsDAC() && !dacOrDacExt.IsDacExtension()) || context.CancellationToken.IsCancellationRequested)
```
SENya1990: There is no need to check for null - the `is` operator performs such checks. `null` value won't pass first `if`

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/LegacyBqlField/LegacyBqlFieldAnalyzer.cs
```diff
@@ -25,23 +25,29 @@ internal override void AnalyzeCompilation(CompilationStartAnalysisContext compil
 
 		private async void Analyze(SymbolAnalysisContext context, PXContext pxContext)
 		{
-			if (context.Symbol is INamedTypeSymbol dacFieldType)
+			if (context.Symbol is INamedTypeSymbol dacOrDacExt)
```
SENya1990: @Derrekam the nested sope becames too big. We have a code style rule that indentation depth shouldn't be more than 3:
https://github.com/Acumatica/Acuminator/blob/dev/docs/dev/CodingGuidelines/CodingGuidelines.md#indentation-depth

I would suggest to reverse big if clause since it's scope is the whole method:
```
if (!(context.Symbol is INamedTypeSymbol dacOrDacExt))
    return;
```
Actually you can merge it with the next if:
```
if (!(context.Symbol is INamedTypeSymbol dacOrDacExt) || (!dacOrDacExt.IsDAC() && !dacOrDacExt.IsDacExtension()) || context.CancellationToken.IsCancellationRequested)
					return;
```

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/LegacyBqlConstant/LegacyBqlConstantFix.cs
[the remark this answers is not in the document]
dropsonic: Ok, I studied this question a little bit further, and it seems that there is no straight answer.
If there is no async code inside the code action itself, we don't know what's best.
`Task.Run` can lead to thread pool starvation, but it heavily depends on the task scheduler that VS uses.
On the other side, `Task.FromResult` can lead to blocking the UI thread since this code action can be executed for multiple pieces of code.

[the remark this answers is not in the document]
SENya1990: I'm not sure if `Task.FromResult` it will lead to hanging of the UI. I have a suspicion that the code fix is already executed asynchrounously. At least the code in BatchFixAllProvider can lead to such conclusion. So I think in the end the `Tas.Run` is not needed.

src/Acuminator/Acuminator.Tests/Tests/StaticAnalysis/LegacyBqlField/Sources/LegacyBqlFieldBad.cs
Derrekam: done

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/LegacyBqlField/LegacyBqlFieldAnalyzer.cs
Derrekam: done

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/LegacyBqlField/LegacyBqlFieldAnalyzer.cs
Derrekam: done

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/LegacyBqlConstant/LegacyBqlConstantFix.cs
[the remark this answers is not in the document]
Derrekam: so should I return ```Task.FromResult``` back?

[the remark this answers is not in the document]
dropsonic: @Derrekam seems so. Sorry for that.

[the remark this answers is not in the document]
SENya1990: @Derrekam Yes, seems that the `FromResult` is alright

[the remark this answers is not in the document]
Derrekam: done

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/LegacyBqlField/LegacyBqlFieldFix.cs
Derrekam: Adding ```PX.Data.BQL``` namespace to a document may lead to some name conflicts in locations that are outside of current code fix scope. For example both ```PX.Data``` and ```PX.Data.BQL``` have ```Minus<TOperand>``` class

SENya1990: Ok, then it will be to expensive to analyze for such conflicts.

PR #181, opened 2019-01-17
ATR-305: do not dig into methods of PXGraph, PXView, and PXCache; new…

metallmonster:
… bypassMethod parameter in the constructor of NestedInvocationWalker

### DIFF
--- a/src/Acuminator/Acuminator.Utilities/Roslyn/NestedInvocationWalker.cs
+++ b/src/Acuminator/Acuminator.Utilities/Roslyn/NestedInvocationWalker.cs
@@ -82,13 +82,7 @@ namespace Acuminator.Utilities.Roslyn
 			else
 			{
 				var pxContext = new PXContext(_compilation);
-				var typesToBypass = new[]
-				{
-					pxContext.PXGraph.Type,
-					pxContext.PXView.Type,
-					pxContext.PXCache.Type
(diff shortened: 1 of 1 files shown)


src/Acuminator/Acuminator.Utilities/Roslyn/NestedInvocationWalker.cs
```diff
@@ -58,14 +59,40 @@ public abstract class NestedInvocationWalker : CSharpSyntaxWalker
...
             CancellationToken = cancellationToken;
 
+			if (bypassMethod != null)
+			{
+				_bypassMethod = bypassMethod;
+			}
+			else
+			{
+				var pxContext = new PXContext(_compilation);
+				var typesToBypass = new[]
```
SENya1990: @metallmonster maybe it makes sense to refactor out creation of forbidden types register - put it into separate field or generate in a virtual method. This way the inheritors of nested walker will have the ability to extend the set of types to which they should not descend

metallmonster: Ok

metallmonster:
@SENya1990 one more try

metallmonster:
@dropsonic please merge if everything is ok

PR #184, opened 2019-01-21
Bugfix/atr 290

metallmonster:


src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/PXGraph/PXGraphSemanticModel.cs
```diff
@@ -41,6 +41,15 @@ public class PXGraphSemanticModel
 		public ImmutableDictionary<string, ActionHandlerInfo> ActionHandlersByNames { get; }
 		public IEnumerable<ActionHandlerInfo> ActionHandlers => ActionHandlersByNames.Values;
 
+		/// <summary>
+		/// Action handlers which are declared in a graph represented by an instance of the class.
+		/// Use this property for diagnostics of graph action handlers
+		/// </summary>
+		public IEnumerable<ActionHandlerInfo> DeclaredActionHandlers => Type == GraphType.PXGraphExtension ?
```
SENya1990: @metallmonster, f I understand correctly this method should return only action handlers declared in a current Symbol. What about standard inheritance of graphs? For example SOInvoiceEntry is inherited from ARInvoiceEntry. This logic will return action handlers from both graphs for SOInvoiceEntry.

metallmonster: @SENya1990 yes, it should return only handlers from the current symbol. I will think about this situation

metallmonster: @SENya1990 I have corrected the code, please review

PR #190, opened 2019-01-23
Feature/atr 55

metallmonster:


src/Acuminator/Acuminator.Analyzers/StaticAnalysis/DacPropertyAttributes/DacPropertyAttributesAnalyzer.cs
```diff
@@ -68,10 +67,48 @@ private static Task AnalyzePropertyAsync(SymbolAnalysisContext symbolContext, PX
...
+		private static void CheckForPXDBCalcedAndUnboundTypeAttributes(SymbolAnalysisContext symbolContext, PXContext pxContext,
+			FieldTypeAttributesRegister fieldAttributesRegister, IPropertySymbol propertySymbol)
+		{
+			symbolContext.CancellationToken.ThrowIfCancellationRequested();
+
+			var attributes = propertySymbol.GetAttributes();
+			var hasUnboundTypeAttribute = false;
+			var hasPXDBCalcedAttribute = false;
+
+			foreach (var attr in attributes)
```
SENya1990: @metallmonster this is incomplete validation. It doesn't consider that some DAC property attributes could be derived from agregators which specify unbound type attribute. The whole point of the `GetFieldTypeAttributesInfos` method is to retrieve the attributes with the consideration for possible aggregators in inheritance chain

SENya1990: ok

metallmonster:
@SENya1990 one more try

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/DacPropertyAttributes/DacPropertyAttributesAnalyzer.cs
```diff
@@ -67,28 +67,28 @@ private static Task AnalyzePropertyAsync(SymbolAnalysisContext symbolContext, PX
...
+			List<(AttributeData Attribute, List<FieldTypeAttributeInfo> Infos)> attributesWithInfos)
 		{
 			symbolContext.CancellationToken.ThrowIfCancellationRequested();
 
-			var attributes = propertySymbol.GetAttributes();
 			var hasUnboundTypeAttribute = false;
 			var hasPXDBCalcedAttribute = false;
 
-			foreach (var attr in attributes)
+			foreach (var (attribute, info) in attributesWithInfos)
```
SENya1990: it should be `infos` instead of `info` because there are multiple DTOs

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/DacPropertyAttributes/DacPropertyAttributesAnalyzer.cs
```diff
@@ -67,28 +67,28 @@ private static Task AnalyzePropertyAsync(SymbolAnalysisContext symbolContext, PX
...
 			{
-				if (attr.AttributeClass.Equals(pxContext.FieldAttributes.PXDBCalcedAttribute))
+				if (attribute.AttributeClass.Equals(pxContext.FieldAttributes.PXDBCalcedAttribute))
```
SENya1990: @metallmonster this validation is not valid. The `attribute` is the same attribute data that you received from the property.
You should not validate it directly, instead you need to use `info` DTO. There is `Kind` property there which will help to find out if the attribute is PXDBCalced. 
To find out if the attribute is field type attribute and is DB unbound you would have to either modify DTO to add this info and `FieldTypeAttributesRegister`

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/DacPropertyAttributes/DacPropertyAttributesAnalyzer.cs
metallmonster: ok

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/DacPropertyAttributes/DacPropertyAttributesAnalyzer.cs
metallmonster: @SENya1990 I agree with the first part of your comment. I'm already using `FieldTypeAttributesRegister` to check whether an attribute is an unbound type or not

metallmonster: @SENya1990 I have looked into the source code and have understood your idea.

PR #200, opened 2019-02-28
 ATR-337: methods for AcuminatorSuppressionMerge tool

metallmonster:


src/Acuminator/Acuminator.Utilities/DiagnosticSuppression/SuppressMessage.cs
```diff
@@ -81,5 +81,33 @@ public override bool Equals(object obj)
...
-	}
+
+        public int CompareTo(SuppressMessage other)
+        {
+            if (Equals(other))
+            {
+                return 0;
+            }
+
+            var idComparison = string.Compare(Id, other.Id);
```
dropsonic: I think that comparison should be ordinal, case-sensitive.
`String.Compare` with two parameters uses the current culture to perform the comparison (see [the docs](https://docs.microsoft.com/en-us/dotnet/api/system.string.compare?view=netframework-4.7.2#System_String_Compare_System_String_System_String_)).

dropsonic:
@metallmonster congrats on having #200 pull request! :)

src/Acuminator/Acuminator.Utilities/DiagnosticSuppression/SuppressMessage.cs
metallmonster: @dropsonic I agree

metallmonster:
@dropsonic one more try

PR #221, opened 2019-04-16
ATR-355: `GetBoundAttributeType()` detect  `IsDBField == true` case i…

saloponov:
…n nested from system attributes.


src/Acuminator/Acuminator.Utilities/Roslyn/PXFieldAttributes/AttributeInformation.cs
```diff
@@ -50,12 +52,15 @@ public AttributeInformation(PXContext pxContext)
 			_context = pxContext;
 
 			var boundBaseTypes = GetBoundBaseTypes(_context);
+			var typesContainsIsDBField = GetTypesContainsIsDBField(_context);
+
 			BoundBaseTypes = boundBaseTypes.ToImmutableHashSet();
+			typesContainsIsDBField.ToImmutableHashSet();
```
metallmonster: The immutable hash set is not assigned to any field or variable

src/Acuminator/Acuminator.Utilities/Roslyn/PXFieldAttributes/AttributeInformation.cs
```diff
@@ -183,6 +195,10 @@ public BoundType GetBoundAttributeType(AttributeData attribute)
 			{
 				var isDbPropertyAttributeArgs = attribute.NamedArguments.Where(arg => IsDBField.Equals(arg.Key, StringComparison.OrdinalIgnoreCase)).ToList();    //case insensitive check
 
+				if (isDbPropertyAttributeArgs.Count == 0 && // IsDBField property defined in base class 
+					TypesContainsIsDBField.Any(typesContainsIsDBField => IsAttributeDerivedFromClass(attribute.AttributeClass, typesContainsIsDBField)))
```
metallmonster: Can you please move linq query to a local variable to make the if condition more readable

src/Acuminator/Acuminator.Tests/Acuminator.Tests.csproj
```diff
@@ -423,6 +426,7 @@
     <EmbeddedResource Include="Tests\StaticAnalysis\ViewDeclarationOrder\Sources\ViewDeclarationOrderWithGraphInheritance.cs" />
     <EmbeddedResource Include="Tests\StaticAnalysis\ViewDeclarationOrder\Sources\ViewDeclarationOrderWithGraphExtension.cs" />
     <EmbeddedResource Include="Tests\StaticAnalysis\ViewDeclarationOrder\Sources\ViewDeclarationOrderWithGraphInheritance_OneCache.cs" />
+    <EmbeddedResource Include="Tests\Utilities\AttributeInformation\Sources\PropertyIsBoundFileFromExternalAssembly.cs" />
```
dropsonic: Naming: what file? why "external" assembly?

src/Acuminator/Acuminator.Tests/Helpers/EmbeddedFileDataAttribute.cs
```diff
@@ -3,15 +3,20 @@
...
-using System.Text;
-using System.Threading.Tasks;
 using System.Runtime.CompilerServices;
 using Xunit.Sdk;
-using Acuminator.Utilities;
 using Acuminator.Utilities.Common;
 
 namespace Acuminator.Tests.Helpers
 {
+	public class EmbeddedFileData2Attribute : EmbeddedFileDataAttribute
```
dropsonic: This is not needed anymore.

src/Acuminator/Acuminator.Tests/Helpers/EmbeddedFileDataAttribute.cs
```diff
@@ -62,14 +69,28 @@ protected EmbeddedFileDataAttribute(string[] fileNames, string testFilePath, str
 				}
 			}
 
+			if (!assemblies.IsNullOrEmpty())
```
dropsonic: Why don't you include PX.Objects as any other dll?

src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/Symbols/FieldAttributeSymbols.cs
```diff
@@ -10,6 +10,9 @@ public class FieldAttributeSymbols
 		/// </summary>
 		private const string PXDBPackedIntegerArrayAttributeFullName_Acumatica2018R2 = "PX.Data.PXDBPackedIntegerArrayAttribute";
 
+		private const string PeriodIDAttribute_Acumatica2019R1 = "PX.Objects.GL.PeriodIDAttribute";
```
dropsonic: Why "Acumatica2019R1"? Isn't it the same in any other versions?
Also, if we move to string literals in all places, it doesn't make sense to extract them to separate constants - it decreases overall readability.

src/Acuminator/Acuminator.Utilities/Roslyn/PXFieldAttributes/AttributeInformation.cs
```diff
@@ -183,6 +195,13 @@ public BoundType GetBoundAttributeType(AttributeData attribute)
...
 				var isDbPropertyAttributeArgs = attribute.NamedArguments.Where(arg => IsDBField.Equals(arg.Key, StringComparison.OrdinalIgnoreCase)).ToList();    //case insensitive check
 
+				
+				var isDBFieldInBaseAcumaticaAttribute = TypesContainsIsDBField.Any(t =>
+					t != null && IsAttributeDerivedFromClass(attribute.AttributeClass, t));
+
+				if (isDbPropertyAttributeArgs.Count == 0 && isDBFieldInBaseAcumaticaAttribute) // IsDBField property defined in base class 
+					return BoundType.DbBound;
+
 				if (isDbPropertyAttributeArgs.Count != 1)  //rare case when there are multiple different "IsDBField" considered
```
SENya1990: This check can be put before the calculation of `isDBFieldInBaseAcumaticaAttribute` because you do not need to know `isDBFieldInBaseAcumaticaAttribute` to make this check.
This is also true for the next check:
```csharp
if (!(isDbPropertyAttributeArgs[0].Value.Value is bool isDbPropertyAttributeArgument))
					return BoundType.Unknown;  //if there is null or values of type other than bool then we don't know if attribute is bound
```

src/Acuminator/Acuminator.Utilities/Common/Constants.cs
```diff
@@ -0,0 +1,81 @@
...
+	{
+		internal static class Types
+		{
+
+			#region FieldAttributeSymbols
+
+			/// <summary>
+			/// The PXDBPackedIntegerArrayAttribute type full name in Acumatica 2018R2. Doesn't exist in previous versions.
+			/// </summary>
+			internal static readonly string PXDBPackedIntegerArrayAttributeFullName_Acumatica2018R2 = "PX.Data.PXDBPackedIntegerArrayAttribute";
```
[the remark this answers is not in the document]
saloponov: Inlining used only for function. This optimization allows instead of calling an assembler call and saving the current state of all processor registers to the stack, to execute a sequence of compiled instructions.

Strings in terms of assembler: working with `strings` based on address arithmetic. There are no differences between `const` or `static readonly`. In first case data saves in metadata and value loads in memory after loading module, static member address saves in MethodTable data structure (https://stackoverflow.com/questions/33562199/static-class-memory-allocation-where-it-is-stored-c-sharp)

https://stackoverflow.com/questions/3674288/whats-the-different-between-ldsfld-and-ldstr-in-il
https://sharplab.io/#v2:C4LglgNgNAJiDUAfAAgBgATIIwG4CwAUMgMyYBM6AwugN6HoOanIAs6AsgBQCUt9jAsADtg6ALYBDYTzoEB8xpQD2QgM5KIAUwB0AeQCuwbQHUATmGCaAMsM2dla4BJGrtAFQCeAB02uAyk7AYADGAGJgmhAw3PhyCvIO6lp6hibmljZCdolOLu7evto5zsDhkdGx8egAkMgA7Oi4AvwKAL4t6O0EXYTClqZCEhCYWABs5FQqqrnAqoSygiKaA0Mj48gUnj5zcQwL8X3Lg8PY46aaEjAqEB4jGAESQWERUegAvOgARAAKABraABFHhJtH8AQAhb4SYIAa00MAAkksAObLACCplMEg8aOAwHMACNDJpPpUDksVsNglNRNgMMURGVXh8fv8gU5Qb8IVDYfCkZZUaYMVicXjCcTSYQOl1WkA===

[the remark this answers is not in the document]
saloponov: @SENya1990 , ok)   for saving 2 operations in initializations static field I will remove `static readonly`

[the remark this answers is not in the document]
saloponov: @SENya1990 , I think that `static readonly` is better than `const`.  
https://www.stum.de/2009/01/14/const-strings-a-very-convenient-way-to-shoot-yourself-in-the-foot/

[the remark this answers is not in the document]
SENya1990: @saloponov  that's why I mentioned the "shoot in a foot" scenario exactly in my first comment and marked it as a very unlikely. 
The inlining may be not precise term because it usually is used for the optimization when compiler replaces the method call with method's body. For constants the compiler replaces the call to the field with actual string literal, which I mentioned as "inlining". Sometimes people also call it "burn" into assembly, etc
I didn't make any benchmarking about it and actuall can rely on the other articles(https://theburningmonk.com/2010/04/net-tips-using-readonly-vs-const-in-csharp/) 
There should be some overhead at least for the load of the type storing static readonly field, 
I agree that difference is very small but people are complaining about Acuminator perfomance, so imho we should try to use most perfomant solutions if they are not too dirty

src/Acuminator/Acuminator.Utilities/Common/Constants.cs
```diff
@@ -0,0 +1,81 @@
...
+	{
+		public static class Types
+		{
+
+			#region FieldAttributeSymbols
+
+			/// <summary>
+			/// The PXDBPackedIntegerArrayAttribute type full name in Acumatica 2018R2. Doesn't exist in previous versions.
+			/// </summary>
+			static readonly string PXDBPackedIntegerArrayAttributeFullName_Acumatica2018R2 = "PX.Data.PXDBPackedIntegerArrayAttribute";
```
SENya1990: @saloponov  all code using Types does not compile because the default protection level when it is not specified explicitly is `private`. Also there is a common style convention in .Net to always specify protection level

src/Acuminator/Acuminator.Utilities/Common/Constants.cs
```diff
@@ -0,0 +1,81 @@
+﻿using System;
```
SENya1990: @saloponov the Common folder is not a very good place for constant file - it was designed to store common helpers for common .net classes and interfaces.
The good place will be imho a separate folder Constants in Roslyn folder or at root level. Or at least move file to the "Roslyn" folder

src/Acuminator/Acuminator.Utilities/Common/Constants.cs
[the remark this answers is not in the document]
SENya1990: After a lot of thought I think it doesn't matter too much what to use - the perfomance hit is minimal,
if XUnit analyzers does it this way then we can too.

src/Acuminator/Acuminator.Utilities/Roslyn/PXFieldAttributes/AttributeInformation.cs
```diff
@@ -66,6 +71,13 @@ public AttributeInformation(PXContext pxContext)
 				context.FieldAttributes.PXDBDataLengthAttribute,
 			};
 
+		private static HashSet<ITypeSymbol> GetTypesContainsIsDBField(PXContext context) =>
```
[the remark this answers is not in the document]
SENya1990: As we discussed earlier the actual way to go is by using the dictionary which will store bool values. So this remark should is not needed

src/Acuminator/Acuminator.Utilities/Roslyn/PXFieldAttributes/AttributeInformation.cs
```diff
@@ -37,9 +38,10 @@ public class AttributeInformation
 		private readonly INamedTypeSymbol _eventSubscriberAttribute;
 		private readonly INamedTypeSymbol _dynamicAggregateAttribute;
 		private readonly INamedTypeSymbol _aggregateAttribute;
-        private readonly INamedTypeSymbol _defaultAttribute;
+		private readonly INamedTypeSymbol _defaultAttribute;
 
 		public ImmutableHashSet<ITypeSymbol> BoundBaseTypes { get; }
+		public ImmutableHashSet<ITypeSymbol> TypesContainsIsDBField { get; }
```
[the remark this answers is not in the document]
saloponov: I have written a simple detection attribute contains `IsDBField = false` excluding case when in inheritance `IsDBField = false` is the latest class.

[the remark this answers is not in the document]
saloponov: `				var isDBFieldSetInFalseInBaseAttribute = TypesContainingIsDBField.Any(t => 
					t.Key != null && t.Value == false && IsAttributeDerivedFromClass(attribute.AttributeClass, t.Key));

				var isDBFieldSetInTrueInBaseAttribute = TypesContainingIsDBField.Any(t =>
					t.Key != null && t.Value == true && IsAttributeDerivedFromClass(attribute.AttributeClass, t.Key));

				if (isDbPropertyAttributeArgs.Count == 0)
				{
					if (isDBFieldSetInFalseInBaseAttribute) // IsDBField = false property defined in base Acumatica class
						return BoundType.Unbound;
					else if (isDBFieldSetInTrueInBaseAttribute) // IsDBField = true property defined in base Acumatica class 
						return BoundType.DbBound;
				}`

src/Acuminator/Acuminator.Utilities/Roslyn/PXFieldAttributes/AttributeInformation.cs
```diff
@@ -183,11 +195,25 @@ public BoundType GetBoundAttributeType(AttributeData attribute)
...
 				if (isDbPropertyAttributeArgs.Count != 1)  //rare case when there are multiple different "IsDBField" considered
 					return BoundType.Unknown;
 
-				if (!(isDbPropertyAttributeArgs[0].Value.Value is bool isDbPropertyAttributeArgument))
-					return BoundType.Unknown;  //if there is null or values of type other than bool then we don't know if attribute is bound
+				var isDBFieldSetInFalseInBaseAttribute = TypesContainingIsDBField.Any(t => 
+					t.Key != null && t.Value == false && IsAttributeDerivedFromClass(attribute.AttributeClass, t.Key));
+
+				var isDBFieldSetInTrueInBaseAttribute = TypesContainingIsDBField.Any(t =>
+					t.Key != null && t.Value == true && IsAttributeDerivedFromClass(attribute.AttributeClass, t.Key));
```
SENya1990: @saloponov 
1) The dictionary can't have a null in its keys so the check `t.Key != null` is redundant
2) You run quite expensive check `IsAttributeDerivedFromClass` twice. You can just write it like this:
```csharp
var typeFromRegister = TypesContainingIsDBField.Keys.FirstOrDefault(t => IsAttributeDerivedFromClass(attribute.AttributeClass, t);
 bool? isDbFieldFromBase = typeFromRegister != null
    ? TypesContainingIsDBField[t]
    : (bool?) null;

if (isDbPropertyAttributeArgs.Count == 0 && isDbFieldFromBase.HasValue)
{	
      return isDbFieldFromBase.Value 
          ? BoundType.DbBound
          : BoundType.Unbound;				
}

//Rest of code
```

src/Acuminator/Acuminator.Tests/Helpers/EmbeddedFileDataAttribute.cs
```diff
@@ -61,10 +59,11 @@ protected EmbeddedFileDataAttribute(string[] fileNames, string testFilePath, str
 						throw new ArgumentException("File name cannot be empty", nameof(internalCodeFileName));
 				}
 			}
-
+			
 			_prefix = GetPrefixFromTestFilePath(testFilePath);
 
 			_fileNames = externalCodeFileNames.IsNullOrEmpty() ? fileNames :  fileNames.Concat(externalCodeFileNames).ToArray();
+
```
metallmonster: Why do you need an empty line before closing brace?

src/Acuminator/Acuminator.Tests/Tests/Utilities/AttributeInformation/AttributeInformationTests.cs
```diff
@@ -137,6 +136,14 @@ private async Task IsBoundAttributeAsync(string source, List<BoundType> expected
 								  new List<bool> { false, false, true },
 								  new string[] { externalAttribute1, externalAttribute2 });
 
+		[Theory]
+		[EmbeddedFileData(@"PropertyIsDBBoundFieldInheritedFromPXObjects.cs")]
+		public Task IsDBFieldFromPXObjectsAsync(string source) =>
+			IsDBFieldPropertyAsync(source,
+									expected: new List<bool> {true, false, true, false},
+									code: new string[] { });
```
metallmonster: I think it is safe to remove this line, am I correct?

src/Acuminator/Acuminator.Utilities/Roslyn/PXFieldAttributes/AttributeInformation.cs
```diff
@@ -181,17 +193,35 @@ public BoundType GetBoundAttributeType(AttributeData attribute)
...
+						? BoundType.DbBound
+						: BoundType.Unbound;
+
+				}
+				else
+				{
+					ITypeSymbol typeFromRegister = TypesContainingIsDBField.Keys.FirstOrDefault(t => IsAttributeDerivedFromClass(attribute.AttributeClass, t));
+					bool? isDbFieldFromBase = typeFromRegister != null
+						? TypesContainingIsDBField[typeFromRegister]
+						: (bool?)null;
```
metallmonster: Why can't you assign `null` instead of `(bool?)null`?

SENya1990: @metallmonster it's because C# compiler can't determine the type of conditional expression without it. It shows compile error if you assign null.

metallmonster: Ok

src/Acuminator/Acuminator.Tests/Helpers/EmbeddedFileDataAttribute.cs
saloponov: Removed

src/Acuminator/Acuminator.Tests/Tests/Utilities/AttributeInformation/AttributeInformationTests.cs
saloponov: Removed.

PR #222, opened 2019-04-22
Feature/atr 367 dev list of diagnostics

dropsonic:


k-s-popova:
@dropsonic , so now we should update the summary file with each new diagnostic, am I correct?

dropsonic:
> @dropsonic , so now we should update the summary file with each new diagnostic, am I correct?

Yes. Basically, it is a copy-paste of a table row from the files describing a diagnostic. Unfortunately, Markdown cannot generate such complex documents automatically.

PR #225, opened 2019-04-26
Bugfix/atr 370 dev

SENya1990:
Merge after ATR-364

### DIFF
--- a/src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/PXGraph/Utils/GraphSymbolUtils.cs
+++ b/src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/PXGraph/Utils/GraphSymbolUtils.cs
@@ -21,7 +21,7 @@ namespace Acuminator.Utilities.Roslyn.Semantic.PXGraph
 		/// <returns>
 		/// The graph from graph extension.
 		/// </returns>
-		public static INamedTypeSymbol GetGraphFromGraphExtension(this ITypeSymbol graphExtension, PXContext pxContext)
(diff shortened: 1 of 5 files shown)


src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/PXGraph/Utils/GraphSymbolUtils.cs
```diff
@@ -40,11 +40,9 @@ public static INamedTypeSymbol GetGraphFromGraphExtension(this ITypeSymbol graph
 				return null;
 
 			ITypeSymbol firstTypeArg = graphExtTypeArgs.Last();
-
-			if (!(firstTypeArg is INamedTypeSymbol pxGraph) || !pxGraph.IsPXGraph())
-				return null;
-
-			return pxGraph;
+			return firstTypeArg.IsPXGraph()
+				? firstTypeArg
```
dropsonic: When can `firstTypeArg` be non-`INamedTypeSymbol` type?

SENya1990: @dropsonic in quite a number of cases when it is type parameter. It's almost any generic graph extension in Pure, for  example `ContactAddressGraph<TGraph>`

SENya1990: Actually `IsPXGraph()` method is also not good enough, it needs to detect graph as a type parameter too

PR #241, opened 2019-05-23
Feature/atr 304 dev code supressing

SENya1990:


src/Acuminator/Acuminator.Vsix/Commands/DiagnosticSuppression/ProjectSuppressionFileGenerator.cs
```diff
@@ -0,0 +1,84 @@
...
+
+
+namespace Acuminator.Vsix.DiagnosticSuppression
+{
+	/// <summary>
+	/// A generator of supression file for project.
+	/// </summary>
+	public static class ProjectSuppressionFileGenerator
+	{
+		private const string CustomSuppresionFileSuffix = "Custom";
```
dropsonic: What if a user wants to edit a shared suppression file?

src/Acuminator/Acuminator.Vsix/Commands/DiagnosticSuppression/RoslynDiagnosticService.cs
```diff
@@ -0,0 +1,178 @@
...
+		private static Type GetDiagnosticDataType() =>
+			(from type in typeof(DocumentId).Assembly.GetTypes()
+			 where type.Name == DiagnosticDataTypeName
+			 select type)
+			.FirstOrDefault();
+
+		private static object GetDiagnosticServiceInstance(IComponentModel componentModel, Type diagnosticAnalyzerServiceType)
+		{
+			Type componentModelType = componentModel.GetType();
+			var getServiceMethodInfo = componentModelType.GetMethod(nameof(componentModel.GetService))
```
dropsonic: Relying on Microsoft's internal implementation is very fragile

src/Acuminator/Acuminator.Vsix/Commands/DiagnosticSuppression/ProjectSuppressionFileGenerator.cs
SENya1990: This part is outdated and actually will be changed. We won't implement custom supression file functionality as I though when I wrote this. I' ll delete it later when I start working on add to suppression file command

src/Acuminator/Acuminator.Vsix/Commands/DiagnosticSuppression/RoslynDiagnosticService.cs
SENya1990: Yes, but we don't have a choice unless they make this internal service public. There is no other way. And the access to internal service gives us huge benefit:
- the ability to add diagnostic to suppression
- the ability to write "Report false alert" command in the future

src/Acuminator/Acuminator.Vsix/Commands/DiagnosticSuppression/ProjectSuppressionFileGenerator.cs
SENya1990: This branch contains some outdated code which will be changed in the future. 
It certainly is not a  finished functionality. I want to preserve part of it because of the wrapper to the Roslyn internal diagnostic service and some other parts. And I don't want to clear it up and finish the "add to suppression file" command right now

src/Acuminator/Acuminator.Vsix/Commands/DiagnosticSuppression/RoslynDiagnosticService.cs
SENya1990: Actually may be we can add some unit tests for it in the future. To check that Microsoft didn't change the implementation

src/Acuminator/Acuminator.Vsix/Commands/DiagnosticSuppression/ProjectSuppressionFileGenerator.cs
```diff
@@ -0,0 +1,84 @@
...
+using Acuminator.Vsix.Utilities;
+
+
+
+namespace Acuminator.Vsix.DiagnosticSuppression
+{
+	/// <summary>
+	/// A generator of supression file for project.
+	/// </summary>
+	public static class ProjectSuppressionFileGenerator
```
metallmonster: My opinion is that this class should be neither removed nor moved to Acuminator.Utilities into DiagnosticsSuppression folder

src/Acuminator/Acuminator.Vsix/Commands/DiagnosticSuppression/SuppressDiagnosticCommand.cs
```diff
@@ -0,0 +1,173 @@
...
+			}
+
+			return diagnosticData?.Where(d => IsAcuminatorDiagnostic(d)) ?? Enumerable.Empty<DiagnosticData>();
+		}
+
+		private static bool IsAcuminatorDiagnostic(DiagnosticData diagnosticData) =>
+			diagnosticData.Id.StartsWith(SharedConstants.AcuminatorDiagnosticPrefix);
+
+
+		private void SuppressDiagnosticsOnNode(SyntaxNode syntaxNode, SemanticModel semanticModel,
```
metallmonster: This method doesn't do anything as far as I can see

src/Acuminator/Acuminator.Vsix/Resources/VSIXResource.resx
```diff
@@ -264,4 +264,10 @@
   <data name="CodeMapWindowTitle" xml:space="preserve">
     <value>Code Map (Experimental)</value>
   </data>
+  <data name="DiagnosticSuppression_NoDiagnosticFound" xml:space="preserve">
+    <value>There are no Acuminator diagnostic on this code snippet.</value>
+  </data>
+  <data name="DiagnosticSuppression_MultipleDiagnosticFound" xml:space="preserve">
+    <value>There are multiple Acuminator diagnostics on this code snippet. The Acuminator can't shoose between them.</value>
```
metallmonster: shoose -> choose

src/Acuminator/Acuminator.Vsix/Commands/DiagnosticSuppression/ProjectSuppressionFileGenerator.cs
dropsonic: @SENya1990 so why can't we keep it in a feature branch? I don't see any major conflicts with VS2019 & Roslyn 2 migration. We planned to release 1.6.2, so it is not desired to have some unfinished functionality in the final build.

SENya1990: Well, I didn't want to keep very old branches but I think you actually right

SENya1990:
I'll close the pull request because I agree with @dropsonic  that it is not a good idea to release unfinished functionality and as I see this branch doesn't require too much change after upgrade.

PR #242, opened 2019-05-23
Feature/atr 124 remove dependency on acumatica

saloponov:


src/Acuminator/Acuminator.Vsix/Coloriser/Roslyn/ParsedDocument.cs
```diff
@@ -16,9 +16,6 @@ namespace Acuminator.Vsix.Coloriser
 {
     public class ParsedDocument
     {
-        private static MetadataReference PXDataReference { get; } =
```
saloponov: @SENya1990 Is it important reference?

SENya1990: Yes it is, the coloring won't work without it.

SENya1990: The code and references in VSIX project won't affect the nuget package with analyzers. We can live with the dependencies in them

SENya1990:
@saloponov  please finish pull request for ATR-355 first and merge it into this this pull request.

src/Acuminator/Acuminator.Utilities/Acuminator.Utilities.csproj
```diff
@@ -53,15 +53,6 @@
     <Reference Include="Microsoft.CodeAnalysis.Workspaces, Version=1.2.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
```
SENya1990: @saloponov please finish pull request for ATR-355 first and merge it into this this pull request.

src/Acuminator/Acuminator.Utilities/Roslyn/Constants.cs
```diff
@@ -0,0 +1,641 @@
...
+			internal static readonly string IPXLocalizableList = "PX.Data.IPXLocalizableList";
+			internal static readonly string PXConnectionScope = "PX.Data.PXConnectionScope";
+
+			internal static readonly string PXGraphExtensionInitialize = "Initialize";
+			internal static readonly string PXLongOperationStartOperation = "StartOperation";
+
+			internal static readonly string IImplementType = "PX.Common.IImplement`1";
+			#endregion
+
+			#region AttributeSymbol.cs
```
SENya1990: I don't think we need .cs endings in region names

src/Acuminator/Acuminator.Utilities/Roslyn/Constants.cs
```diff
@@ -0,0 +1,641 @@
...
+			internal static readonly string CustomPredicate = "PX.Data.CustomPredicate";
+			internal static readonly string AreSame2 = "PX.Data.AreSame`2";
+			internal static readonly string AreDistinct2 = "PX.Data.AreDistinct`2";
+			internal static readonly string Required1 = "PX.Data.Required`1";
+			internal static readonly string Argument1 = "PX.Data.Argument`1";
+			internal static readonly string Optional1 = "PX.Data.Optional`1";
+			internal static readonly string Optional2 = "PX.Data.Optional2`1";
+			internal static readonly string BqlCommand = "PX.Data.BqlCommand";
+			internal static readonly string IBqlParameter = "PX.Data.IBqlParameter";
+			//internal static readonly string PXSelectBase1 = "PX.Data.PXSelectBase`1";//dublicate
```
SENya1990: You can remove the duplicate

src/Acuminator/Acuminator.Utilities/Roslyn/Constants.cs
```diff
@@ -0,0 +1,641 @@
...
+				internal static readonly string FieldVerifying2 = "PX.Data.Events+FieldVerifying`2";
+				internal static readonly string FieldUpdating2 = "PX.Data.Events+FieldUpdating`2";
+				internal static readonly string FieldUpdated2 = "PX.Data.Events+FieldUpdated`2";
+				internal static readonly string CommandPreparing2 = "PX.Data.Events+CommandPreparing`2";
+				internal static readonly string ExceptionHandling2 = "PX.Data.Events+ExceptionHandling`2";
+
+			}
+			#endregion
+
+			#region ExceprionSymbols.cs
```
SENya1990: mistype : Excep**t**ionSymbols 
The `.cs` ending is redundant

src/Acuminator/Acuminator.Utilities/Roslyn/Constants.cs
```diff
@@ -0,0 +1,641 @@
...
+			public static readonly string IBqlOrderBy = "PX.Data.IBqlOrderBy";
+			public static readonly string IBqlSortColumn = "PX.Data.IBqlSortColumn";
+			public static readonly string IBqlFunction = "PX.Data.IBqlFunction";
+			public static readonly string IBqlPredicateChain = "PX.Data.IBqlPredicateChain";
+			public static readonly string IBqlTable = "PX.Data.IBqlTable";
+
+			#endregion
+
+
+			#region Template
```
SENya1990: Why do you need the TemplateAttributeOrClass  or TemplateDelegates? They are not referenced in code

src/Acuminator/Acuminator.Utilities/Roslyn/Constants.cs
```diff
@@ -0,0 +1,641 @@
...
+			{
+				internal static readonly string SetList = "SetList";
+			}
+
+			#endregion
+
+
+		}
+
+		internal sealed class GenerateConstStringAttribute : Attribute
```
SENya1990: Whar is the purpose of this attribute?

dropsonic:
@SENya1990 @metallmonster please review

src/Acuminator/Acuminator.Utilities/Roslyn/Constants.cs
dropsonic: I agree with @SENya1990.
@metallmonster In this case, it is done intentionally to keep all string literals in one place instead of separating them into several files.

metallmonster: @dropsonic what are the benefits of this approach? For me it is difficult to work with a big file full of regions. I don't know, maybe it is Ok for you, but I'd prefer to have several files unified through single namespace. It would improve readability and decrease complexity of the file.

dropsonic: @metallmonster because from the user standpoint it should be the same class (`Types` and `TypeNames`) instead of multiple files. It could be achieved by using partial classes, but I don't think it makes any sense because nobody will read this file.

SENya1990: Well actually I would also prefer if this file was splitted into multiple partial files - a partial Types class + nested class with constants. I think it will be more structured than regions - it will be easier to add new constant in a right place.

SENya1990: But I won't insist on it because imho it isn't too important issue

SENya1990: There are too many duplicates and unobvious nested types for constants which were part of legacy and etc. TypeNames, for example. We could either keep them as they are and postpone refactoring - in this case in my opinion one file is actually better. Or we can regroup this constants in scope of this pull request and in this case I would prefer a few big logically groups of constants each kept in a separate file.

metallmonster: I agree with @SENya1990 , if we don't want to refactor this file it is OK to keep it like this. But later we need to do something with it.

saloponov:
@dropsonic  @SENya1990  @metallmonster , review new changes, please. This version classified string constants by
a. TypeNames - `Acuminator.Utilities.Roslyn.Constants.TypeNames` 
b. TypeFullNames - `Acuminator.Utilities.Roslyn.Constants.TypeFullNames`
c. EventsNames - `Acuminator.Utilities.Roslyn.Constants.EventsNames` 
d. EventArgsNames - `Acuminator.Utilities.Roslyn.Constants.EventArgsNames`
e. DelegateNames - `Acuminator.Utilities.Roslyn.Constants.DelegateNames`

Please will write to me, If you see errors classify in strings. An example of this approach can be found in `xunit.analyzer` https://github.com/xunit/xunit.analyzers/blob/master/src/xunit.analyzers/Analysis/CoreContext.cs

saloponov:
@SENya1990 @metallmonster  You can find an example: https://github.com/xunit/xunit.analyzers/blob/cebd67a5c3534e43a2e7b5fa0367d9ed1260bb3c/src/xunit.analyzers/Constants.cs

SENya1990:
They have  a quite strange approach to the access modifiers. I'd guess this is someone's personal code style cause it is definitely not a common practice in .Net to omit access modifiers or specify `internal` class with `internal` members.

PR #243, opened 2019-05-24
Feature/upgrade/atr 383 roslyn upgrade

SENya1990:


src/Acuminator/Acuminator.Tests/Tests/Utilities/NestedInvocationWalker/NestedInvocationWalkerTests.cs
```diff
@@ -159,7 +165,7 @@ public async Task Constructor(string text)
 			walker.Locations.Should().BeEquivalentTo((Line: 13, Column: 14));
 		}
 
-		[Theory]
+		[Theory(Skip = "The test is incorrect for current implementation of nested walker which does not step inside lambdas")]
 		[EmbeddedFileData("LocalLambda.cs")]
```
SENya1990: @dropsonic please check out this unit test, it is incorrectly written. The exception walker, as it is implemented now, will never visit any nodes in lambda expressions because the base nested walker skips them

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/EventHandlers/EventHandlerAnalyzer.cs
```diff
@@ -75,17 +75,17 @@ protected override void AnalyzeSymbol(SymbolAnalysisContext context, PXContext p
...
 
-		private void AnalyzeLambda(OperationAnalysisContext context, PXContext pxContext)
-		{
-			if (context.Operation is ILambdaExpression lambdaExpression)
-			{
-				var symbolAnalysisContext =
-					new SymbolAnalysisContext(lambdaExpression.Signature, context.Compilation, context.Options,
-											  context.ReportDiagnostic, d => true, // this check is covered inside context.ReportDiagnostic
-											  context.CancellationToken);
+		//private void AnalyzeLambda(OperationAnalysisContext context, PXContext pxContext)
```
metallmonster: Why did you comment these lines?
I propose delete them if they are unnecessary.

src/Acuminator/Acuminator.Utilities/Properties/AssemblyInfo.cs
```diff
@@ -1,30 +0,0 @@
-﻿using System.Resources;
```
metallmonster: What happened with AssemblyInfo file?

src/Acuminator/Acuminator.Analyzers/Properties/AssemblyInfo.cs
```diff
@@ -1,20 +0,0 @@
-﻿using System.Reflection;
```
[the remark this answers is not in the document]
SENya1990: @metallmonster  in net standard it is not required - it is embedded in csproj
https://montemagno.com/how-to-convert-a-pcl-library-to-net-standard-and-keep-git-history/

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/EventHandlers/EventHandlerAnalyzer.cs
SENya1990: In new Roslyn this API is not found. I think we should later look where this API went and check if it is deprecated or just has a different name

src/Acuminator/Acuminator.Utilities/Properties/AssemblyInfo.cs
SENya1990: in net standard it is not required - it is embedded in csproj
https://montemagno.com/how-to-convert-a-pcl-library-to-net-standard-and-keep-git-history/

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/EventHandlers/EventHandlerAnalyzer.cs
metallmonster: Ok, but I think we need to figure it out before we will merge this branch to be sure that we will merge stable branch

SENya1990: This pull request goes into internal branch `feature/ATR-382-dev-VS-UPGRADE` which I created as a feature branch to work on upgrade. It will be merged after the 1.6.2 release. 
The AnalyzeLambda API wasn't used before in Acuminator - it was added by @dropsonic  and as far as I remember it required some sort of switch in Roslyn or didn't worked with Visual Studio. I remember that it was some kind of experimental feature in the version of Roslyn that we used.
It was stored to be turned on after the upgrade of the Roslyn

metallmonster: Ok, in that case we need to figure it out before merging it to the `dev` branch

dropsonic: This part was not "experimental", it covers some valid cases with lambdas. But indeed we need to look into changes in the new Roslyn API.

src/Acuminator/Acuminator.Tests/Tests/Utilities/NestedInvocationWalker/NestedInvocationWalkerTests.cs
dropsonic: If it is incorrect, why did it pass before? It requires a valid location from the nested walker.

src/Acuminator/Acuminator.Vsix/Commands/Formatter/BqlFormatter.cs
```diff
@@ -54,5 +52,20 @@ public static BqlFormatter FromTextView(IWpfTextView textView)
 
 			return new BqlFormatter(newLineCharacter, !convertTabsToSpaces, tabSize, indentSize);
 		}
+
+		private static IEnumerable<SyntaxTrivia> GetUseTabsModeIndentTrivias(int indentSize, int tabSize)
```
dropsonic: What is this fix and how it is related to the migration?

SENya1990: This fix was needed due to removal of Prepend/Append collection helpers in Acuminator.Utilities. 
Somehow this methods are included in Linq version used by .Net Standart but not included into .Net 4.7.2 which is used in VSIX.  Therefore there was a compilation error in Net standard due to ambigous method call - so I removed our helpers but there was a call to Prepend here.
I didn't want to move just one utility method to VSIX because it was used only in this place so I created this method

src/Acuminator/Acuminator.Tests/Tests/Utilities/NestedInvocationWalker/NestedInvocationWalkerTests.cs
SENya1990: @dropsonic  that's a very interesting question. Actually the syntax tree created by the old version of Roslyn is quite strange:
--Class
  --Method
     --Body
         1. LocalDeclarationStatement
         2. ThrowStatement
         3. ExpressionStatement

The Syntax tree in the new Roslyn however looks like this:
[(https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzhgHwAEAmARgFgAoIgBgAIiyUBuauxsgOgBkBLAHYBHNjQZMuAFRgIMojkwCso6gOwBbGLgAO2MDHoBBMAFd1g7BmjUA3tQCQRAMyMS9AGIQIDu1XuOXIhR6ABEIAGUITQwAC0EAcwAKAEoHe19/ewA3bCh6PQx6AF56FOKAPnpYqAgAd3oBGHqAUQR9bQw+CAEU0UyC3rSAXwcRqiGgA===)]

--Class
  --Method
     --Body
         1. LocalDeclarationStatement
         2. ExpressionStatement

I suspect it is caused by the inroduction of the throw expressions to C# 7
You can see now the reason why the test passed

PR #250, opened 2019-06-21
Feature/atr 304 dev code supression 2.0

SENya1990:
merge after atr 382 branch


SENya1990:
@k-s-popova please review the following strings:
Submenu for suppression commands:
`Suppress Acuminator diagnostic`

Command name:
`In Supression File`   (so it will be read like this: Suppress Acuminator diagnostic -> In Supression File)

Error messages:
`Failed to add diagnostic to the suppression file. Check if the file {0} is accessible.`
`Failed to find the suppression file for project {0}.`
`There are multiple Acuminator diagnostics on this code snippet. The Acuminator can't shoose between them.`
`There are no Acuminator diagnostic on this code snippet.`

SENya1990:
@metallmonster , @dropsonic 
I have fixed all bugs that appeared during manual testing, plese review.
Note, that unfixed bug ATR-391 affects the suppression command functionality too. 
If you change the solution the suppression files internal collection is not collected and the suppression command can't find a suppression file to write to

dropsonic:
@SENya1990 `Failed to find the suppression file for project {0}.` - shouldn't it create the file if it doesn't exist?

SENya1990:
@dropsonic  Yes, and it does so. This is an error message in case the command fails to create the suppression file:
```csharp
var (suppressionFileRoslynDoc, project) = await GetProjectAndSuppressionFileAsync(diagnostic.ProjectId);
bool suppressionFileExists = suppressionFileRoslynDoc != null;

if (!suppressionFileExists)
{
	if (project == null)
		return;

	SuppressionFile suppressionFile = SuppressionManager.CreateSuppressionFileForProject(project);
	suppressionFileExists = suppressionFile != null;
}

if (!suppressionFileExists || 
	!SuppressionManager.SuppressDiagnostic(semanticModel, diagnostic.Id, diagnostic.DataLocation.SourceSpan.Value,
										   diagnostic.DefaultSeverity, Package.DisposalToken))
{
	ShowErrorMessage(suppressionFileRoslynDoc, project);
}
```

k-s-popova:
Change as follows:
In Supression File -> in Suppression File (small i and double p)

Error messages:
The diagnostic cannot be added to the suppression file. Check if the {0} file is accessible.
The suppression file for the {0} project cannot be found.
There are multiple Acuminator diagnostics in this code snippet. Acuminator can suppress only one diagnostic at a time.
There are no Acuminator diagnostics in this code snippet.

SENya1990:
After a discussion with @k-s-popova I'll rename the command:
In Supression File -> in Acuminator Suppression File

SENya1990:
Changed strings to reviewed ones

src/Acuminator/Acuminator.Utilities/DiagnosticSuppression/SuppressionFile.cs
```diff
@@ -78,17 +76,19 @@ internal static bool IsSuppressionFile(string path)
 				messages = LoadMessages(fileSystemService, suppressionFilePath);
 			}
 
-			var fileWatcher = fileSystemService.CreateWatcher(suppressionFilePath);
+			ISuppressionFileWatcherService fileWatcher;
 
+			lock (fileSystemService)
```
dropsonic: You already have a lock on `fileSystemService` in the call stack above the current method

dropsonic:
If there are two diagnostics on the same symbol, they suppressed simultaneously (which is not obvious for the user).

SENya1990:
@dropsonic no, in this case the error message will be shown and nothing will be done.
The case for multiple diagnostics is handled by this code:
```
protected override Task SupressMultipleDiagnosticOnNodeAsync(List<DiagnosticData> diagnosticData, Document document, SyntaxNode syntaxRoot,
															 SemanticModel semanticModel, SyntaxNode nodeWithDiagnostic)
{
	MessageBox.Show(VSIXResource.DiagnosticSuppression_MultipleDiagnosticFound, AcuminatorVSPackage.PackageName);
	return Task.CompletedTask;
}
```

src/Acuminator/Acuminator.Utilities/DiagnosticSuppression/SuppressionFile.cs
SENya1990: The lock allows recursive calls by the same thread and at the same time we cannot guarantee that only `SuppressionManager` will call the `SuppressionFile`. It's a public class which accepts some file service so I thought that it's better to make sure that the service is accessed exclusively

dropsonic:
@SENya1990 it didn't show any error messages to me and added two diagnostics to the suppression file. But I wasn't able to reproduce it again.
The second time when I added the suppression, VS still showed me the diagnostic (but the suppression file was modified).

SENya1990:
@dropsonic that's a very strange bug. Since this is a temporary solution until we implement the suppression as "code fix" I suggest to ignore it because it is hard to reproduce.

Also keep in mind that VS won't immediately clear diagnostics - they are cached. So for some time it will still show you the diagnostic. The actual time of refresh depends on the size of solution. Big solutions require more time

src/Acuminator/Acuminator.Utilities/DiagnosticSuppression/SuppressionFile.cs
metallmonster: @SENya1990 why do we need a look here?

src/Acuminator/Acuminator.Utilities/DiagnosticSuppression/SuppressionFile.cs
```diff
@@ -126,28 +126,32 @@ private static XElement ElementFromMessage(SuppressMessage message)
...
-
-			return new SuppressMessage(id, target, syntaxNode);
-		}
-
 		public static HashSet<SuppressMessage> LoadMessages(ISuppressionFileSystemService fileSystemService, string path)
 		{
-			var document = fileSystemService.Load(path);
+			XDocument document;
+
+			lock (fileSystemService)
```
metallmonster: @SENya1990 why do we need a look here?

src/Acuminator/Acuminator.Utilities/DiagnosticSuppression/SuppressionManager.cs
```diff
@@ -72,6 +54,19 @@ public void ReloadFile(object sender, SuppressionFileEventArgs e)
 			_fileByAssembly[assembly] = newFile;
 		}
 
+		private (SuppressionFile File, string Assembly) LoadFileAndTrackItsChanges(string suppressionFilePath, bool generateSuppressionBase)
+		{
+			lock (_fileSystemService)
```
metallmonster: @SENya1990 why do we need a look here?

src/Acuminator/Acuminator.Utilities/DiagnosticSuppression/SuppressionManager.cs
```diff
@@ -88,53 +83,82 @@ public static void Init(ISuppressionFileSystemService fileSystemService, IEnumer
 
 		public static void SaveSuppressionBase()
 		{
-			if (Instance == null)
+			CheckIfInstanceIsInitialized();
+			
+			lock (Instance._fileSystemService)
```
metallmonster: @SENya1990 why do we need a look here?

src/Acuminator/Acuminator.Utilities/DiagnosticSuppression/SuppressionManager.cs
```diff
@@ -88,53 +83,82 @@ public static void Init(ISuppressionFileSystemService fileSystemService, IEnumer
...
+			project.ThrowOnNull(nameof(project));
+			CheckIfInstanceIsInitialized();
 
-			var (assembly, message) = GetSuppressionInfo(semanticModel, diagnostic, cancellation);
+			//First check if file already exists to dismiss threads withou acquiring the lock
+			if (Instance._fileByAssembly.TryGetValue(project.Name, out var existingSuppressionFile) && existingSuppressionFile != null)
+				return existingSuppressionFile;
 
-			if (assembly == null)
+			lock (Instance._fileSystemService)
```
metallmonster: @SENya1990 why do we need a look here?

src/Acuminator/Acuminator.Utilities/DiagnosticSuppression/SuppressionManager.cs
```diff
@@ -88,53 +83,82 @@ public static void Init(ISuppressionFileSystemService fileSystemService, IEnumer
...
-				return false;
+				//Second check inside the lock if file already exists 
+				Instance._fileByAssembly.TryGetValue(project.Name, out existingSuppressionFile);
+				return existingSuppressionFile ?? AddNewSuppressionFileImpl();
 			}
 
-			var file = _fileByAssembly.GetOrAdd(assembly, (SuppressionFile)null);
-
-			if (file == null)
+			//---------------------------------------------Local Function--------------------------------------------------
```
metallmonster: @SENya1990 from my point of view if a file has code regions or a function has comments as separators it should be splitted into parts without regions/comments-separators;

src/Acuminator/Acuminator.Utilities/DiagnosticSuppression/SuppressionManager.cs
```diff
@@ -88,53 +83,82 @@ public static void Init(ISuppressionFileSystemService fileSystemService, IEnumer
...
+				string suppressionFilePath = Path.Combine(projectDir, suppressionFileName);
 
-			if (file.GenerateSuppressionBase)
-			{
-                if (diagnostic?.Descriptor.DefaultSeverity != DiagnosticSeverity.Info)
-                {
-                    file.AddMessage(message);
-                }
+				//Create new xml document and get its text
+				System.Xml.Linq.XDocument newXDocument = SuppressionFile.NewDocumentFromMessages(Enumerable.Empty<SuppressMessage>());
```
metallmonster: Why not to add the System.Xml.Linq namespace to usings?

src/Acuminator/Acuminator.Utilities/DiagnosticSuppression/SuppressionManager.cs
```diff
@@ -88,53 +83,82 @@ public static void Init(ISuppressionFileSystemService fileSystemService, IEnumer
...
+				string docText = GetXDocumentStringWithDeclaration(newXDocument);
 
-				return true;
+				//Add file to project and hard drive
+				var roslynSuppressionFile = project.AddAdditionalDocument(suppressionFileName, docText, filePath: suppressionFilePath);
+
+				if (!project.Solution.Workspace.TryApplyChanges(roslynSuppressionFile.Project.Solution))
```
metallmonster: Why do we need this check?

src/Acuminator/Acuminator.Utilities/DiagnosticSuppression/SuppressionManager.cs
```diff
@@ -88,53 +83,82 @@ public static void Init(ISuppressionFileSystemService fileSystemService, IEnumer
...
 				return false;
-			}
 
-			return true;
+			var (fileAssemblyName, suppressMessage) = SuppressMessage.GetSuppressionInfo(semanticModel, diagnosticID, 
+																						 diagnosticSpan, cancellation);
+			if (fileAssemblyName.IsNullOrWhiteSpace() || !suppressMessage.IsValid)
+				return false;
+
+			lock (Instance._fileSystemService)
```
metallmonster: Why do we need this lock?

src/Acuminator/Acuminator.Utilities/DiagnosticSuppression/SuppressionManager.cs
```diff
@@ -146,13 +170,16 @@ public static IEnumerable<SuppressionDiffResult> ValidateSuppressionBaseDiff()
 
 			var diffList = new List<SuppressionDiffResult>();
 
-			foreach (var entry in Instance._fileByAssembly)
+			lock (Instance._fileSystemService)
```
metallmonster: Why do we need this lock?

src/Acuminator/Acuminator.Utilities/DiagnosticSuppression/SuppressionManager.cs
```diff
@@ -185,85 +212,63 @@ private static SuppressionDiffResult CompareFiles(SuppressionFile oldFile, Suppr
...
-				.FirstOrDefault();
-
-			// Use first variable in case of field declaration as it may contain multiple variables and therefore
-			// cannot be used to obtain declared symbol
-			return targetNode is FieldDeclarationSyntax fieldDeclaration ?
-				fieldDeclaration.Declaration?.Variables[0] :
-				targetNode;
+			var builder = new System.Text.StringBuilder(capacity: 65);
+
+			using (TextWriter writer = new Utf8StringWriter(builder))
```
metallmonster: The class Utf8StringWriter is used only in this place. Maybe it is better to remove it?

src/Acuminator/Acuminator.Utilities/DiagnosticSuppression/SuppressionFile.cs
SENya1990: @metallmonster because the VS throws exception about access to file from another thread.

src/Acuminator/Acuminator.Utilities/DiagnosticSuppression/SuppressionFile.cs
SENya1990: @metallmonster because the VS throws exception about access to file from another thread.

src/Acuminator/Acuminator.Utilities/DiagnosticSuppression/SuppressionManager.cs
SENya1990: @metallmonster because the VS throws exception about access to file from another thread.

src/Acuminator/Acuminator.Utilities/DiagnosticSuppression/SuppressionManager.cs
SENya1990: @metallmonster because the VS throws exception about access to file from another thread.
So I added exclusive access to all places where the file service is used to access the file

src/Acuminator/Acuminator.Utilities/DiagnosticSuppression/SuppressionManager.cs
SENya1990: @metallmonster because the VS throws exception about access to file from another thread.

src/Acuminator/Acuminator.Utilities/DiagnosticSuppression/SuppressionManager.cs
SENya1990: It's just one place - didn't want to litter Intellisense

src/Acuminator/Acuminator.Utilities/DiagnosticSuppression/SuppressionFile.cs
SENya1990: @metallmonster because the VS throws exception about access to file from another thread.

src/Acuminator/Acuminator.Utilities/DiagnosticSuppression/SuppressionFile.cs
SENya1990: @metallmonster because the VS throws exception about access to file from another thread.

src/Acuminator/Acuminator.Utilities/DiagnosticSuppression/SuppressionManager.cs
SENya1990: We need to call the TryApplyChanges to change the solution. This method is the one actually creating a new file on disk.
So it can fail and this check returns null if it failed.

src/Acuminator/Acuminator.Utilities/DiagnosticSuppression/SuppressionManager.cs
metallmonster: But it is not as readable as it could be. Maybe it is better to use "var" instead of "System.Xml.Linq.XDocument"?

src/Acuminator/Acuminator.Utilities/DiagnosticSuppression/SuppressionManager.cs
SENya1990: This is just stylistic issue for me - there is only one local function and I like that extra separation to be able to quickly see where the main function ends. I can live without it but don't see much problem here.  On the other hand imho making it a separate private function has following flaws:
1) all parameters should be passed explicitly
2) wider encapsulation scope - this local function should be called only from that `CreateSuppressionFileForProject` method
3) Placement - imho is is easier to read code sequentially from top to bottom. But some guidelines like StyleCop advice you to put public methods first and private methods with the implementation are below. I think the local method can deal with this issue. 

Overall - it's just separator to see where the main function ends. Is it that bad?

src/Acuminator/Acuminator.Utilities/DiagnosticSuppression/SuppressionManager.cs
SENya1990: @metallmonster because the VS throws exception about access to file from another thread

src/Acuminator/Acuminator.Utilities/DiagnosticSuppression/SuppressionManager.cs
SENya1990: @metallmonster because the VS throws exception about access to file from another thread

src/Acuminator/Acuminator.Utilities/DiagnosticSuppression/SuppressionManager.cs
metallmonster: Ok. I understand that. No, it is not so bad, just my comment, lets keep it as is. Lets discuss it in detail if we add something about local functions to the code style.

src/Acuminator/Acuminator.Utilities/DiagnosticSuppression/SuppressionManager.cs
SENya1990: The problem solved here is that that it's hard to get xml string with a xml declaration from XDocument. You need to use code like this to get it. And the string writer doesn't allow you to set custom encoding for string writer. You need to created derived class to do it. And its default one is UTF-16.
 That's why `Utf8StringWriter` is created. You need a class derived from string writer.

src/Acuminator/Acuminator.Utilities/DiagnosticSuppression/SuppressionManager.cs
SENya1990: ok. 
```suggestion
				var newXDocument = SuppressionFile.NewDocumentFromMessages(Enumerable.Empty<SuppressMessage>());
```

src/Acuminator/Acuminator.Utilities/DiagnosticSuppression/SuppressionManager.cs
SENya1990: Yes, I think we already have something there.

src/Acuminator/Acuminator.Utilities/DiagnosticSuppression/SuppressionManager.cs
metallmonster: And why can't we use UTF-16?

SENya1990: The standard generated xml file with `xdocument.Save(path)` is UTF-8. I think it is not good if sometimes the suppression file will have UTF-8 and other times - UTF-16. I'd like to stick to the default implementation provided by the XDocument

PR #251, opened 2019-06-24
ATR-401: added missing nuget packages

SENya1990:


src/NuGet.Config
```diff
@@ -0,0 +1,6 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
```
dropsonic: Do we need this?

SENya1990: It is autogenerated now. I tried to delete it but it was re-generated on build

SENya1990: I suppose it is part of "new format" for csproj file when you use embedded nuget references instead of package.config.

SENya1990:
Try to build now

PR #252, opened 2019-06-25
Bugfix/atr 399 dac extension default attribute analyzer fails with argument null exception

saloponov:


saloponov:
Merge after ATR-401

PR #256, opened 2019-07-11
Bugfix/atr 408 dev

SENya1990:


src/Acuminator/Acuminator.Vsix/Resources/VSIXResource.resx
```diff
@@ -294,4 +294,10 @@
   <data name="PackageLoad_WaitMessage" xml:space="preserve">
     <value>Loading the Acuminator package</value>
   </data>
+  <data name="FailedToLoadTheSuppressionFile" xml:space="preserve">
+    <value>Acuminator failed to load the suppression file. The suppression mechanism will be disabled.</value>
```
SENya1990: @k-s-popova please verify the error message for incorrect suppression file. The message consists of two parts:
"Acuminator failed to load the suppression file. The suppression mechanism will be disabled."
"The details of error: {0}."
The {0} is placeholder for the exception message. The final message should look like this:
"Acuminator failed to load the suppression file. The suppression mechanism will be disabled.
The details of error:
_EXCEPTION DETAILS_"

k-s-popova: @SENya1990, please correct as follows:
Acuminator cannot load the suppression file. The suppression mechanism will be turned off.
Error details: {0}
Can a user fix the error somehow? Should we add something like "If you have fixed the error, you can turn on the suppression mechanism in the Acuminator settings." or it will be turned on automatically if the error is fixed?

k-s-popova: @SENya1990 , if the error is always because of bad XML structure, I would change the message a bit:
 "Acuminator cannot load the suppression file because of an error in the XML structure of the file. The suppression mechanism will be turned off.
Error details: {0}"
I would still say something about how to turn it on again.

dropsonic: @SENya1990 I agree with @k-s-popova. How can a user turn it on again (~ to reload the suppression file)?
A typical scenario is when the suppression file is corrupted, the user fixed it and wants to reload it again.

SENya1990: @dropsonik, @k-s-popova this sounds like a separate feature. I suspect that because there is a watcher on the file, you can fix it and it will be reloaded. But I need to test it.

SENya1990: @k-s-popova the error depends on the type of exception. We can show different error messages for bad xml structure, security error (like access denied) or general IO error (failed to load file)
Do you think it will be better this way?
If so could you please provide the messages for these errors?

k-s-popova: @SENya1990 , no, let's stay with the first variant then:
Acuminator cannot load the suppression file. The suppression mechanism will be turned off.
Error details: {0}

SENya1990: @k-s-popova, @dropsonic  I just checked and the behaviour is great - you don't need to restart VS, just fix the file.  Every time the mistake happens the error will be shown to user, so the behaviour is consistent. 
@k-s-popova  could you please provide a final version of the error message conidering this behaviour? Maybe we can add something like - fix the file?

k-s-popova: @SENya1990 ,
Acuminator cannot load the suppression file. The suppression mechanism will be turned off.
If you fix the error, the suppression mechanism will be turned on automatically.
Error details: {0}

SENya1990: изображение

SENya1990: Here is the message dialog. Fixed strings

SENya1990:
@k-s-popova, @dropsonic  please approve

PR #258, opened 2019-07-14
Feature/atr 331 dev dac aggregating analyzer

SENya1990:


src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/Base/OverridableItemsCollection.cs
```diff
@@ -0,0 +1,74 @@
...
+		{
+			infoConstructor.ThrowOnNull(nameof(infoConstructor));
+			TWriteableInfo info = infoConstructor(rawData, declarationOrder);
+			Add(info);
+		}
+
+		internal void Add<TWriteableInfo>(TWriteableInfo info)
+		where TWriteableInfo : TInfo, IWriteableBaseItem<TInfo>
+		{
+			if (info?.Name == null)
```
metallmonster: I suggest to throw an exception instead of return

src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/Dac/Enums/DacType.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Acuminator.Utilities.Roslyn.Semantic.Dac
+{
+	public enum DacType
+	{
+		None,
```
metallmonster: Why do we need this value?

src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/Dac/Utils/DacPropertySymbolUtils.cs
```diff
@@ -0,0 +1,273 @@
...
+		/// <param name="dacPropertyInfos">The DAC property infos.</param>
+		/// <param name="dacOrDacExtension">The DAC or DAC extension.</param>
+		/// <param name="startingOrder">The declaration order which should be assigned to the first DTO.</param>
+		/// <returns/>
+		private delegate int AddDacPropertyInfoWithOrderDelegate<TInfo>(OverridableItemsCollection<TInfo> dacPropertyInfos,
+																		ITypeSymbol dacOrDacExtension, int startingOrder)
+		where TInfo : IOverridableItem<TInfo>;
+
+
+		#region Dac Properties
```
metallmonster: I propose to remove this region and split file if it is not readable without the region

src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/Dac/Utils/DacPropertySymbolUtils.cs
```diff
@@ -0,0 +1,273 @@
...
+		/// <param name="cancellation">(Optional) Cancellation token.</param>
+		/// <returns>The DAC property symbols with nodes from DAC.</returns>
+		public static OverridableItemsCollection<DacPropertyInfo> GetDacPropertiesFromDac(this ITypeSymbol dac, PXContext pxContext,
+																						  bool includeFromInheritanceChain = true,
+																						  CancellationToken cancellation = default)
+		{
+			pxContext.ThrowOnNull(nameof(pxContext));
+
+			if (!dac.IsDAC(pxContext))
+				return new OverridableItemsCollection<DacPropertyInfo>();
```
metallmonster: Why not to throw error here?

src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/Dac/Utils/DacPropertySymbolUtils.cs
```diff
@@ -0,0 +1,273 @@
...
+		public static OverridableItemsCollection<DacPropertyInfo> GetPropertiesFromDacOrDacExtensionAndBaseDac(this ITypeSymbol dacOrExtension,
+																											   PXContext pxContext,
+																											   CancellationToken cancellation = default)
+		{
+			pxContext.ThrowOnNull(nameof(pxContext));
+
+			bool isDac = dacOrExtension.IsDAC(pxContext);
+
+			if (!isDac && !dacOrExtension.IsDacExtension(pxContext))
+				return new OverridableItemsCollection<DacPropertyInfo>();
```
metallmonster: Why not to throw error here?

src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/Dac/Utils/DacPropertySymbolUtils.cs
```diff
@@ -0,0 +1,273 @@
...
+		/// The DAC fields from DAC.
+		/// </returns>
+		public static OverridableItemsCollection<DacFieldInfo> GetDacFieldsFromDac(this ITypeSymbol dac, PXContext pxContext, 
+																				   bool includeFromInheritanceChain = true,
+																				   CancellationToken cancellation = default)
+		{
+			dac.ThrowOnNull(nameof(dac));
+			pxContext.ThrowOnNull(nameof(pxContext));
+
+			if (!dac.IsDAC(pxContext))
```
metallmonster: Why not to throw error here?

src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/Dac/Utils/DacSymbolsHierarchyUtils.cs
```diff
@@ -0,0 +1,170 @@
...
+                return Enumerable.Empty<ITypeSymbol>();
+            }
+
+            var extensionBaseType = dacExtension.BaseType;
+            var typeArguments = extensionBaseType.TypeArguments;
+            var dacType = typeArguments.LastOrDefault();
+
+            if (dacType == null || !dacType.IsDAC(pxContext))
+            {
+                return Enumerable.Empty<ITypeSymbol>();
```
metallmonster: Why not to throw error here?

src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/Base/OverridableItemsCollection.cs
SENya1990: Ok, agree

src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/Dac/Enums/DacType.cs
SENya1990: To be similar to PXGraph + some a value to represent "not a DAC". I think may be it will be used in the future. Or maybe not. Let's remove it for now

src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/Dac/Utils/DacPropertySymbolUtils.cs
SENya1990: I agree

src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/Dac/Utils/DacPropertySymbolUtils.cs
SENya1990: Well I think there are two approaches with utility methods when there is an invalid input:
1) throw an exception
2) do nothing and quickly return.

There are examples of both in Acuminator and in frankly for this class I copied the logic from the similar helper for PXGraph Actions. So, one part of the answer is - to be consistent with the helpers for graph semantic model.

But also I'm not sure that first approach is always the best if we are not passing something completely invalid like null reference. 
This is an utility method so it is expected to be called in different contexts. Maybe in one place there should be an exception thrown. But the caller can throw it by himself if needed. 
On the other hand, if someone needs a silent behaviour then he is left with no choice but use try/catch.
Maybe the best approach is to use something like TryXXX pattern which will also return a flag of success to allow the caller to distinguish between failure and successfull empty result. But it is not very convenient or elegant to use such helper. 

Overall there are some pros and cons for different approaches. I suggest to stay with a "silent" one to be consistent with graph semantic utils and most of existing utils. 
If we really want to throw error then it should be consistent, so other helpers should be rewritten too.

src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/Dac/Utils/DacPropertySymbolUtils.cs
SENya1990: See my reply above

src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/Dac/Utils/DacPropertySymbolUtils.cs
SENya1990: See my reply above

src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/Dac/Utils/DacSymbolsHierarchyUtils.cs
SENya1990: See my reply above

PR #278, opened 2019-08-09
ATR-391: reinitialize SuppressionManager after a solution was opened

metallmonster:


src/Acuminator/Acuminator.Vsix/AcuminatorVSPackage.cs
```diff
@@ -244,25 +249,32 @@ protected override void Dispose(bool disposing)
...
+        }
+
+        private async System.Threading.Tasks.Task InitializeSuppressionManagerAsync()
+        {
+            var workspace = await this.GetVSWorkspaceAsync();
+            var additionalFiles = workspace.CurrentSolution.Projects
+                .SelectMany(p => p.AdditionalDocuments)
+                .Select(d => new SuppressionManagerInitInfo(d.FilePath, generateSuppressionBase: false));
+
+            SuppressionManager.Init(new SuppressionFileSystemService(), additionalFiles);
```
SENya1990: @metallmonster everything looks fine. However, there is one thing bugging me.
 If I understand correctly you create a new suppression manager during the re-opening of the solution with its own file service and file watcher. 
Suppose we open for example same solution, or another solution but with the same projects.
The old file service will still perform an operation on a suppression file while a new one will try to access it. I'm not sure if this is entirely possible but still I want to discuss it. Maybe we need to lock in the old fileservice to wait until it finishes its job.
@dropsonic what do you think about the possibility?

dropsonic: @metallmonster @SENya1990 a user can open two instances of Visual Studio that look at the same project, so I think it should be possible to edit this file simultaneously.

metallmonster: done

src/Acuminator/Acuminator.Utilities/DiagnosticSuppression/SuppressionManager.cs
```diff
@@ -73,15 +73,20 @@ public static void Init(ISuppressionFileSystemService fileSystemService, IEnumer
...
-			if (Instance != null)
-			{
-				throw new InvalidOperationException($"{typeof(SuppressionManager).Name} has been already initialized");
-			}
-
-			Instance = new SuppressionManager(fileSystemService, suppressionFiles);
-		}
-
-		public static void SaveSuppressionBase()
+            if (Instance?._fileSystemService != null)
```
dropsonic: It might be possible that between `if` and `lock` statements another thread will initialize `Instance` property, so we need to double-check if `Instance?._fileSystemService` is null inside the lock.

metallmonster: @dropsonic it might be so, but the logic inside `if` is fired if `Instance?._fileSystemService` is not null

SENya1990: Well, now I thnik that monitor surely prolong life of locked object and it won't be GCed inside lock. So it should be safe, however, the double lock technique is quite standard - I don't thnik that it will hurt to add it for extra safety.

PR #279, opened 2019-08-21
ATR-129: analyzer and codefix to detect public PX classes without xml…

metallmonster:
… comment description


src/Acuminator/Acuminator.Analyzers/Resources.resx
```diff
@@ -159,14 +159,11 @@
   <data name="PX1006Title" xml:space="preserve">
     <value>This order of view declarations will cause creation of one cache instance for the following DACs: {0} and {1}</value>
   </data>
-  <data name="PX1007Fix" xml:space="preserve">
-    <value>Create a DAC instance by using the factory method</value>
-  </data>
-  <data name="PX1007MessageFormat" xml:space="preserve">
-    <value>An instance of {0} must be created with the PXCache&lt;{0}&gt;.CreateInstance() factory method</value>
+  <data name="PX1007FixExcludeClass" xml:space="preserve">
+    <value>Exclude class from API Reference</value>
```
metallmonster: @k-s-popova please review these 3 messages

k-s-popova: Exclude the class from the API Reference

src/Acuminator/Acuminator.Analyzers/Resources.resx
```diff
@@ -159,14 +159,11 @@
...
   </data>
   <data name="PX1007Title" xml:space="preserve">
-    <value>A DAC instance must be created with the PXCache.CreateInstance() factory method</value>
+    <value>A public class should have description in the summary tag of the XML annotation</value>
```
k-s-popova: A public class should have a description in the summary XML tag

src/Acuminator/Acuminator.Analyzers/Resources.resx
```diff
@@ -480,4 +477,7 @@
   <data name="PX1043TitleRowPersistedNonISV" xml:space="preserve">
     <value>Only the methods of the PXCache.Persist family can be used to save changes to the database from a RowPersisted event handler</value>
   </data>
+  <data name="PX1007FixAddDescription" xml:space="preserve">
+    <value>Add description</value>
```
k-s-popova: Add a description

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PublicClassXmlComment/PublicClassXMLCommentAnalyzer.cs
```diff
@@ -0,0 +1,172 @@
...
+            var isAcumaticaClass = IsInAcumaticaRootNamespace(classDeclarationSyntax, syntaxContext.SemanticModel, syntaxContext.CancellationToken);
+            if (!isAcumaticaClass)
+            {
+                return;
+            }
+
+            CheckXmlComment(syntaxContext, classDeclarationSyntax);
+        }
+
+        private bool IsInAcumaticaRootNamespace(ClassDeclarationSyntax classDeclarationSyntax, SemanticModel semanticModel,
```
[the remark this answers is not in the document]
dropsonic: Everything else looks awesome!

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PublicClassXmlComment/PublicClassXMLCommentAnalyzer.cs
```diff
@@ -0,0 +1,172 @@
...
+        private void AnalyzeClassDeclaration(SyntaxNodeAnalysisContext syntaxContext)
+        {
+            syntaxContext.CancellationToken.ThrowIfCancellationRequested();
+
+            if (!(syntaxContext.Node is ClassDeclarationSyntax classDeclarationSyntax))
+            {
+                return;
+            }
+
+            var isPublicClass = classDeclarationSyntax.Modifiers.Any(SyntaxKind.PublicKeyword);
```
SENya1990: there is a more complex scenario of a public nested class in an internal outer calss (or even some nesting hierarchy).
I think we need to check not only this type but also its containing types that all of them are public.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PublicClassXmlComment/PublicClassXMLCommentAnalyzer.cs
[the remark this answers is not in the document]
SENya1990: I agree - this should be a helper for ITypeSymbol.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PublicClassXmlComment/PublicClassXMLCommentAnalyzer.cs
```diff
@@ -0,0 +1,172 @@
...
+
+            CheckXmlComment(syntaxContext, classDeclarationSyntax);
+        }
+
+        private bool IsInAcumaticaRootNamespace(ClassDeclarationSyntax classDeclarationSyntax, SemanticModel semanticModel,
+            CancellationToken cancellationToken)
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+
+            var classDeclarationSymbol = semanticModel.GetDeclaredSymbol(classDeclarationSyntax, cancellationToken);
```
SENya1990: Since you're using typesymbol anyway I would obtain it earlier - before the check for public modifier.
First of all, you can quit earlier in case of null. Second - the check is easier, just check
`classDeclarationSymbol.DeclaredAccessibility`

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PublicClassXmlComment/PublicClassXMLCommentAnalyzer.cs
```diff
@@ -0,0 +1,172 @@
...
+            if (xmlComment == null)
+            {
+                ReportDiagnostic(syntaxContext, classDeclarationSyntax, FixOption.NoXmlComment);
+				return;
+            }
+
+			var excludeTag = xmlComment
+				.ChildNodes()
+				.OfType<XmlEmptyElementSyntax>()
+				.Where(s => XmlCommentExcludeTag.Equals(s?.Name?.ToString(), StringComparison.Ordinal))
```
SENya1990: A little nitpicking so I won't insist on it but Where and FirstOrDefault can be combined in one:
```csharp
FirstOrDefault(s => XmlCommentExcludeTag.Equals(s?.Name?.ToString(), StringComparison.Ordinal))
```

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PublicClassXmlComment/PublicClassXMLCommentAnalyzer.cs
```diff
@@ -0,0 +1,172 @@
...
+				.FirstOrDefault();
+			if (excludeTag != null)
+			{
+				return;
+			}
+
+			var summaryTag = xmlComment
+                .ChildNodes()
+                .OfType<XmlElementSyntax>()
+                .Where(n => XmlCommentSummaryTag.Equals(n?.StartTag?.Name?.ToString(), StringComparison.Ordinal))
```
SENya1990: the same as the previous comment

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PublicClassXmlComment/PublicClassXMLCommentAnalyzer.cs
metallmonster: It was done on purpose. I'm prefer such chain of Linq methods in which every method performs single operation. In this particular case it clearly indicates where we filter results (`Where`) and where we control what element of filtered results we want to use (`FirstOrDefault`).

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PublicClassXmlComment/PublicClassXmlCommentFix.cs
```diff
@@ -0,0 +1,215 @@
...
+		}
+
+		private SyntaxNode AddXmlCommentDescription(SyntaxNode rootNode, ClassDeclarationSyntax classDeclarationSyntax,
+			string description, CancellationToken cancellation)
+        {
+            cancellation.ThrowIfCancellationRequested();
+
+			var xmlDescriptionTrivia = Trivia(
+				DocumentationComment(
+					XmlSummaryElement(
```
SENya1990: Most of the documentation tools as well as Visual Studio generate summary with this pattern:
```xml
<summary>
Description
</summary>
```
So, I think we also need to stick with it instead of:
```xml
<summary>Description</summary>
```

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PublicClassXmlComment/PublicClassXMLCommentAnalyzer.cs
SENya1990: Ok, I won't insist on it

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PublicClassXmlComment/PublicClassXMLCommentAnalyzer.cs
metallmonster: ok

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PublicClassXmlComment/PublicClassXMLCommentAnalyzer.cs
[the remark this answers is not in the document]
metallmonster: ok

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PublicClassXmlComment/PublicClassXMLCommentAnalyzer.cs
metallmonster: ok

metallmonster: @SENya1990 done

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PublicClassXmlComment/PublicClassXMLCommentAnalyzer.cs
metallmonster: @SENya1990 done

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PublicClassXmlComment/PublicClassXmlCommentFix.cs
metallmonster: @SENya1990 done

PR #296, opened 2019-10-07
Feature/atr 400 dev suppress in file as code fix

SENya1990:


SENya1990:
The strings were verified by @k-s-popova

src/Acuminator/Acuminator.Utilities/DiagnosticSuppression/IO/Base/I_IOErrorProcessor.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+using System.Runtime.CompilerServices;
+
+namespace Acuminator.Utilities.DiagnosticSuppression.IO
+{
+	/// <summary>
+	/// Interface for i/o error processing.
+	/// </summary>
+	public interface I_IOErrorProcessor
```
dropsonic: inconsistent naming

SENya1990: Hm and how it should be called? I just don't like the `IIOErrorProcessor` because of double I

dropsonic: but what's wrong with double I? `IInternable`, `IInvoice`, `IIf` are classical examples

SENya1990: Maybe I do not like it because of four vowels in a row. But ok, I'll fix it

SENya1990: done

PR #301, opened 2019-10-11
Bugfix/atr 461 dev

SENya1990:


SENya1990:
@dropsonic I want to fix issues with BuildAction and with handling deletion of suppression file in separate issues

PR #305, opened 2019-10-16
Bugfix/atr 463 dev px1007 fix

SENya1990:


SENya1990:
merge after atr-468

SENya1990:
@dropsonic yes, I've read about it. However, I don't like this idea - I am forced to turn on/off the unrelated option in order to disable some random diagnostic. What if I'm an internal developer who wants to see more strict checks? Or I can be an OEM developer who disabled this option

dropsonic:
@SENya1990 another solution: we can set Severity = Hidden, and place a ruleset file with Severity = Warning to our projects.

SENya1990:
@dropsonic I like this solution much more. But I think there are several cons. Are you sure that Hidden diagnostics are not displayed in the IDE? Not only in the code editor but also in the error list window? Some people may find it annoying if they are displayed there.  
The second problem in my opinion is that we will have to create and add custom rulesets to all our projects. If for example we add them to Acu2019R2 then this diagnostic won't be shown for earlier versions which are still in use.

dropsonic:
@SENya1990 I'm quite sure about the severity because it is the same mechanism that is used by the ruleset engine. But, indeed, we cannot add ruleset files to already released branches.
We can add a special explicit option for this diagnostic.

SENya1990:
@dropsonic Yes, I thought about that too. The second option is to somehow create a list of assemblies for which this diagnostic can be applied. Unfortunately for it to work with nuget it can only be hard-coded.
And there is actually a third option. Why don't we use IAssemblySymbol.
It has Public Key Token (but there is a problem that some of our projects are not signed) and also attributes which we can read. And in those attributes we can get `AssemblyCompany` attribute and check that its Acumatica.

In any case since this issue has a separate related item ATR-464 I propose to do it in scope of that item

metallmonster:
@SENya1990 @dropsonic I've spoke with ISV certification team, they have not received any ISV solution with PX root namespace so maybe detecting PX namespace is not a bad idea after all.

SENya1990:
I agree with @metallmonster . It seems strange to write a customization code in the namespace used by the library provider. It's the same as if we write in the System namespace. Maybe some of the small partner customizations are written this way but I doubt they use Visual Studio or Acuminator.
We can return to this issue if someone comes with complaint

dropsonic:
@SENya1990 @metallmonster  Acuminator is used not only by ISV developers, so we can't make decisions based on ISV solutions only.

SENya1990:
@dropsonic  but OEM developers and big solutions that have its own namespace usually have their own namespace

dropsonic:
@SENya1990 yes, but our partners and clients have many small solutions created from the Customization Manager which utilizes PX namespace.

dropsonic:
Will be fixed in ATR-464.

PR #311, opened 2019-10-17
ATR-473: changes in PX1007

k-s-popova:


docs/diagnostics/PX1007.md
```diff
@@ -8,18 +8,20 @@ This document describes the PX1007 diagnostic.
 | PX1007 | A public entity or DAC property should have a description in the `summary` XML tag. | Warning | Available | 
 
 ## Diagnostic Description
-Any public class, interface, structure, enumeration, delegate, or their member that is defined in a `PX` namespace (such as the `PX.Objects.AR` namespace) should have a description in the `summary` XML tag. These descriptions can be used for building the API and DAC references for the Acumatica ERP libraries.
+Any public class, interface, struct, enumeration, delegate, or their member that is defined in a `PX` namespace (such as the `PX.Objects.AR` namespace) should have a description in the `summary` XML tag. These descriptions can be used for building the API and DAC references for the Acumatica ERP libraries.
```
SENya1990: We removed the logic checking the namespace. It works for any namespace now

k-s-popova: This is a description of the requirement. We need only the entities from PX namespaces to be documented.
It is written below that the diagnostic works for any namespace.

SENya1990: I can't see the description. For me it looks a bit confusing. For example, I have an alert for class in namespace `MyNamespace`, open the diagnostic description from VS and see that
> Any public class, interface, struct, enumeration, delegate, or their member that is defined in a `PX`
I'm confused now because my class is obviously not defined there

docs/diagnostics/PX1007.md
```diff
@@ -8,18 +8,20 @@ This document describes the PX1007 diagnostic.
 | PX1007 | A public entity or DAC property should have a description in the `summary` XML tag. | Warning | Available | 
 
 ## Diagnostic Description
-Any public class, interface, structure, enumeration, delegate, or their member that is defined in a `PX` namespace (such as the `PX.Objects.AR` namespace) should have a description in the `summary` XML tag. These descriptions can be used for building the API and DAC references for the Acumatica ERP libraries.
+Any public class, interface, struct, enumeration, delegate, or their member should have a description in the `summary` XML tag. The descriptions of the entities and members that are defined in a `PX` namespace (such as the `PX.Objects.AR` namespace) can be used for building the API and DAC references for the Acumatica ERP libraries.
```
SENya1990: @k-s-popova  I'm also not sure about their member part - we currently do this only for public DAC properties. Should we mention other members?

k-s-popova: In my opinion, this is only the current limitation. All public entities and members should have descriptions.

SENya1990: I agree with that, But this is a documentation of diagnostic behavior, not a guide to good  code style. I think for documentation to be actual we should write what the diagnostic actually do, not what we want it to do or what it will do in the future version.

k-s-popova: For all our diagnostics we write the common rule in the beginning. I insist that the common rule is about all entities and members. The limitation about the current work of the diagnostic is described.

SENya1990: ok.

PR #327, opened 2019-12-31
Bugfix/atr 509 dev

SENya1990:


src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PublicClassXmlComment/PublicClassXMLCommentAnalyzer.cs
```diff
@@ -46,8 +49,20 @@ internal enum FixOption
...
+		/// <summary>
+		/// Check that compillation is a unit test assembly. The check is implemented by searching for <c>Test</c> word in the assembly name. 
+		/// It is a common pattern which on one hand is used almost everywhere and on the other hand allows us to distance from the concrete unit test frameworks
+		/// and support not only xUnit but also others like NUnit.
+		/// </summary>
+		/// <param name="compilation">The compilation.</param>
+		/// <returns/>
+		protected virtual bool IsUnitTestAssembly(Compilation compilation) =>
+			compilation?.AssemblyName != null && 
+			compilation.AssemblyName.ToUpperInvariant().Contains(UnitTestAssemblyMarker);
```
dropsonic: I think it should be `EndsWith` because there are some assemblies that don't contain tests itself (e.g., `MyProject.TestFramework`).

SENya1990: I disagree because:
1. Such assemblies usually are still related to the tests and probably are internal so it's not good to force their documentation.
2. The pattern to end everything with `Tests` is not so common as it looks. Sure, most of test assemblies fit into it but there are exceptions:
- in Pure PX.Objects.Tests.Common assembly
- in OEMs tests, for example, Lexware, there are also assemblies `Tests.Common`, "AR.Tests.Utils", "TestBase" and "TestDataBuilder"

dropsonic: Assemblies like `PX.Objects.Tests.Common` contain shared entities that can be used by other developers and OEMs so I think it is a good idea to document what's inside them.
As for Lexware, they should use commonly accepted naming conventions.

SENya1990: It's hard for me to tell anything about `PX.Objects.Tests.Common`. It looks to me that it contains legacy - old attempts to write unit tests for application which were created before the unit test framework. I don't think they should be documented - either deleted or rewritten. 

I have more examples - in net tools there are `PX.Tests.Common` and `PX.Tests.Common.Api`. It is unclear to me if the `PX.Data.GI.TestFramework` should also be documented.
 Also my investigation of platform tests showed another problem - there are unit test assembly `PX.Olap.Benchmarks` with performance tests which do not have 'Test' in the name. Instead there is a "Benchmark". Should we also look for the word "Benchmark"?

Regarding Lexware - they just won't do it as they won't switch from NUnit unless there is a serious issues. They have their own conventions and coding best practices which they sometimes try to enforce on us with approximately same result. Also, they are probably not the only one partner with unit tests for Acumatica. We can't enforce them all. 
Perhaps we can introduce additional mechanism to enable/disable Acuminator analysis via assembly attribute or editor.config/app.config option to make the process more flexible.

There is also a bigger question - do we need to run Acuminator diagnostics at all in unit test projects?

dropsonic: `PX.Data.GI.TestFramework` is better be documented. It was documented on Wiki: https://wiki.acumatica.com/display/TND/Generic+Inquiry+Test+Framework

As for OEMs, we provide rules and best practices. One of the best practices is to use xUnit and commonly accepted naming conventions. Acuminator is one of the ways to force them to accept these best practices (in my opinion).

SENya1990: Still there are two other test assemblies - `PX.Tests.Common` and `PX.Tests.Common.Api`.
Regarding OEMs - I agree that we can provide best practices, this is one of Acuminator goals. But we can't enforce a practice or a pattern that wasn't even formulated and approved by all internal programmers. And, if it was, it wasn't communicated even among out internal developers. How many internal developers are aware of it? Is it documented somewhere?
Moreover, we already have a violation of this pattern in our own test projects naming.

I suggest to think about alternative explicit way to declare that the Acuminator should run on the project which will take priority over the naming rule

PR #339, opened 2020-01-31
Feature/atr 518 dev rn213

k-s-popova:
Please review the release notes and other changes in docs for 2.1.3.


docs/ReleaseNotes.md
```diff
@@ -1,6 +1,25 @@
...
 
+## Acuminator 2.1.3
+Acuminator 2.1.3 includes the bug fixes and enhancements described in this section, as well as the features that have been implemented in previous versions.
+
+### Enhancements
+In Acuminator 2.1.3, the following enhancements have been implemented:
+ - Acuminator now does not perform static analysis of projects whose names contain `Test` or `Benchmark`. Generally, these projects contain tests of an application and do not include any application logic; therefore there is no need to analyze them with Acuminator.
+ - The memory usage, performance, and stability of the code map have been improved.
+ - The synchronization of the code map with the changes in the DACs has been improved.
+ - Now a developer can collapse or expand the selected node in the code map if the developer clicks **Collapse All** or **Expand Add** (respectively) in the context menu for the selected node.
```
SENya1990: @k-s-popova collapse or expand the selected node **with all its descendant nodes**

docs/ReleaseNotes.md
```diff
@@ -1,6 +1,25 @@
...
+ - Multiple changes have been made to the [PX1007](diagnostics/PX1007.md) diagnostic to reduce the number of warnings displayed by the diagnostic. For more information about the situations when the diagnostic is not displayed, see the description of the diagnostic in [PX1007](diagnostics/PX1007.md).
+ - If a developer suppresses the [PX1007](diagnostics/PX1007.md) diagnostic with a comment, the default comment message is now the following: _Acuminator disable once PX1007 NoXmlCommentForPublicEntityOrDacProperty to be documented later_.
```
SENya1990: Do we need the underscores surrounding the message?

docs/ReleaseNotes.md
```diff
@@ -1,6 +1,25 @@
...
+
+### Fixed Bugs
+In this version of Acuminator, the following bugs have been fixed:
+ - The [PX1007](diagnostics/PX1007.md) diagnostic was displayed for partial classes with XML comments on one of the declarations.
+ - The [PX1007](diagnostics/PX1007.md) diagnostic was displayed for class fields of DACs.
```
SENya1990: This section can be added to the previous one:
> - Multiple changes have been made to the [PX1007](diagnostics/PX1007.md) diagnostic to reduce the number of warnings displayed by the diagnostic. For more information about the situations when the diagnostic is not displayed, see the description of the diagnostic in [PX1007](diagnostics/PX1007.md).

Overall, it is also decreasing of the number of PX1007 alerts

docs/ReleaseNotes.md
```diff
@@ -1,6 +1,25 @@
...
+ - The [PX1094](diagnostics/PX1094.md) diagnostic was displayed for classes derived from `PXMappedCacheExtension`.
+ - Acuminator failed on the Acuminator suppression file if the file had an incorrect XML structure.
```
SENya1990: I would add that now Acuminator validates suppression file to have correct xml structure. We created xml schema for suppression file and use it to validate the correctness of structure

docs/diagnostics/PX1007.md
```diff
@@ -15,13 +15,28 @@ The code fix does one of the following:
...
-The diagnostic is currently displayed for all public entities (classes, interfaces, structs, enumerations, and delegates) and DAC property fields no matter in which namespace they are located. The diagnostic is not displayed for other fields, properties, and methods of public entities. 
-
-The diagnostic is not displayed for child classes if their parent class has the `exclude` XML tag.
+The new diagnostic is available only if the value of **Tools > Options > Acuminator > Code Analysis > Enable PX1007 diagnostic** is _True_. By default, the value of this option is _False_. 
+
+This diagnostic is intended for Acumatica ERP developers; however, it can be used by any developer to provide documentation of the source code.
+
+The diagnostic is currently displayed for the following public items:
+ - Classes that are not DACs 
+ - DACs with the `PXCacheName` attribute
```
SENya1990: @k-s-popova this is not correct. The diagnostic is displayed for all public classes without xml comment unless they have one of the attributes: `PXHidden`, `PXInternalUseOnly`, or `Obsolete`. 
In case the class has one of the mentioned attributes the diagnostic won't be shown not only for it but also for all its members suitable for the diagnostic - public nested types, DAC properties.
This attributes work like "exclude" tag.

docs/diagnostics/PX1007.md
```diff
@@ -15,13 +15,28 @@ The code fix does one of the following:
...
+ - DAC property fields
+ - Interfaces
+ - Structs
+ - Enumerations
+ - Delegates
+
+The diagnostic is not displayed for the following items: 
+ - Fields, properties, and methods of public entities (except DAC property fields).
```
SENya1990: I would say type members of public entities unless it is a nested public type (class/struct/enum/interface/delegate type) or a public DAC property
Because there are other kinds of type members - instance and static constructors, events

docs/diagnostics/PX1007.md
```diff
@@ -15,13 +15,28 @@ The code fix does one of the following:
...
+ - Child classes if their parent class has the `exclude` XML tag.
```
SENya1990: The used term is "nested classes" and I would say that exclude tag disables diagnostic for all type members of a type. This way it also includes nested structs/enums/interfaces/delegate types and public DAC properties

docs/diagnostics/PX1007.md
```diff
@@ -15,13 +15,28 @@ The code fix does one of the following:
...
+ - DACs with the `PXHidden`, `PXInternalUseOnly`, or `Obsolete` attribute.
+ - Class fields of DACs. In DACs, XML annotations should be specified only for property fields.
```
SENya1990: I don't think it will hurt to specify xml comment on both class field and property. I would rather not mention this at all

docs/diagnostics/PX1007.md
```diff
@@ -15,13 +15,28 @@ The code fix does one of the following:
...
+ - DAC property fields
+ - Interfaces
+ - Structs
+ - Enumerations
+ - Delegates
+
+The diagnostic is not displayed for the following items: 
+ - Fields, properties, and methods of public entities (except DAC property fields).
+ - Child classes if their parent class has the `exclude` XML tag.
+ - DACs with the `PXHidden`, `PXInternalUseOnly`, or `Obsolete` attribute.
```
SENya1990: No only DACs - everything that is marked with this attributes and in case of types also all their type members

docs/diagnostics/PX1007.md
```diff
@@ -15,13 +15,28 @@ The code fix does one of the following:
...
+ - Class fields of DACs. In DACs, XML annotations should be specified only for property fields.
+ - System fields of DACs, such as `CreatedDateTime` or `LastModifiedDateTime`.
```
SENya1990: I think a full list of fields must be here:
- NoteID
- CreatedByID
- CreatedByScreenID
- CreatedDateTime
- LastModifiedByID
- LastModifiedByScreenID
-  LastModifiedDateTime
-  tstamp

docs/diagnostics/PX1007.md
```diff
@@ -15,13 +15,28 @@ The code fix does one of the following:
...
+ - Public items from the assemblies whose names contain `Test` or `Benchmark`.
```
SENya1990: I think that it is not a specific trait of PX1007 diagnostic - it can be applied to all of them.
Therefore, I would place it in some general place like Summary or DiagnosticSuppression files

docs/ReleaseNotes.md
k-s-popova: added

docs/ReleaseNotes.md
k-s-popova: We usually use them in other our docs...

docs/ReleaseNotes.md
k-s-popova: Do you mean both bullets about PX1007? or only the latest one (about class fields)?

docs/ReleaseNotes.md
k-s-popova: added

docs/diagnostics/PX1007.md
k-s-popova: @SENya1990 , it was in the requirements in the JIRA issue to display the diagnostics for the DACs with PXCacheName, cause we are not interested in other DACs for the DAC reference.

docs/diagnostics/PX1007.md
k-s-popova: modified

docs/diagnostics/PX1007.md
k-s-popova: modified

docs/diagnostics/PX1007.md
k-s-popova: removed

docs/diagnostics/PX1007.md
k-s-popova: modified

docs/diagnostics/PX1007.md
k-s-popova: Summary is definitely not the right place, as well as DiagnosticSuppression. This is written in Release Notes. I can just remove it here.

docs/diagnostics/PX1007.md
k-s-popova: modified

docs/ReleaseNotes.md
SENya1990: ok

docs/ReleaseNotes.md
SENya1990: about class fields. 
The first one is more close to a bug - you already have a comment on a public partial entity, just in the other file. So, it is a false alert, not a change of diagnostic rules

k-s-popova: removed about class fields

docs/diagnostics/PX1007.md
SENya1990: @k-s-popova but the DAC is still a public class in C# and the diagnostic requires all public classes to be documented.  Why should it be excluded? Moreover, having a DAC without either PXCacheName or PXHidden is also an error - we have another diagnostic for it. 
We added support to not show PX1007 for PXHidden. So, it's imho not good to check additionally for PXCacheName. It will be more costly without adding real value

k-s-popova: @SENya1990 , OK, but it was marked with (V) in jira, so that was documented :)
Modified

docs/diagnostics/PX1007.md
SENya1990: I'm not sure - maybe at least it should be in the readme. Or DiagnosticSuppression. 
Imho, it is wrong to read all release notes just to find out description of product behavior because it is missing in all other places. 
This is somewhat important information that some assemblies won't be analyzed. If I was a user of Acuminator I would definitely want to know more about the rules of exclusion

k-s-popova: We don't have full documentation for Acuminator and we do not pretend to write it...

SENya1990: Yes, we don't document the API and most of implementation details but we try to document the functionality and this info is related to functionality.  So, it will be great if we can add a small text about it to the summary

docs/diagnostics/PX1007.md
SENya1990: yes, my bad, should have edited Jira item. Thanks

docs/diagnostics/PX1007.md
k-s-popova: So far we do not document the UI as well (except in Release Notes).
Text about enable code analysis switch and the excluded unit test projects added to Summary.md

SENya1990: Maybe we should write the detailed docs for UI at some point. Anyway, thanks a lot!

PR #357, opened 2020-03-19
Feature/atr 539 dev rn

EAndrosova:
Add 2.2 RN and new diagnostic descriptions

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -1,6 +1,23 @@
 # Acuminator Release Notes
 This document provides information about fixes, enhancements, and key features that are available in Acuminator.
 
+## Acuminator 2.2
+Acuminator 2.2 includes the bug fixes and enhancements described in this section, as well as the features that have been implemented in previous versions.
+
+### Enhancements
+In Acuminator 2.2, the following enhancements have been implemented:
+ - The [PX1019](diagnostics/PX1019.md) diagnostic has been added to check for DAC fields of non strings type with the AutoNumberAttribute.
+ - The [PX1020](diagnostics/PX1020.md) diagnostic has been added to check if the length of a string DAC field with the AutoNumberAttribute is sufficient.
+ - The [PX1022](diagnostics/PX1022.md) diagnostic has been added to check if a DAC or graph extension is declared as a public class.
+ - Context menu of the Acuminator has been updated: sort commands have been divided in the context menu, active sort indicator have been added, menu control templates have been redesigned to integrate the Visual Studio theme.
+ - The following complementary information about DAC fields is displayed in a separate column: field type, whether the field is an identity field, whether the field is bound or unbound.
+ - It is now possible to sort the code map. The context menu now includes the following new commands: sort node's children by declaration, sort node's children by name, sort all node descendants by declaration, sort all node descendants by name.
+
+### Fixed Bugs
+In this version of Acuminator, the following bugs have been fixed:
+ - Acuminator failed on Microsoft Visual Studio 2017.
+ - The Acuminator package was loaded in Visual Studio at random time including loading after the analysis was already executed.
+
 ## Acuminator 2.1.3
 Acuminator 2.1.3 includes the bug fixes and enhancements described in this section, as well as the features that have been implemented in previous versions.
 
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -6,17 +6,17 @@ Acuminator 2.2 includes the bug fixes and enhancements described in this section
 
 ### Enhancements
 In Acuminator 2.2, the following enhancements have been implemented:
- - The [PX1019](diagnostics/PX1019.md) diagnostic has been added to check for DAC fields of non strings type with the AutoNumberAttribute.
- - The [PX1020](diagnostics/PX1020.md) diagnostic has been added to check if the length of a string DAC field with the AutoNumberAttribute is sufficient.
+ - The [PX1019](diagnostics/PX1019.md) diagnostic has been added to check for DAC fields of non-strings types with the `AutoNumberAttribute` attribute.
+ - The [PX1020](diagnostics/PX1020.md) diagnostic has been added to check if the length of a string DAC field with the `AutoNumberAttribute` attribute is sufficient.
  - The [PX1022](diagnostics/PX1022.md) diagnostic has been added to check if a DAC or graph extension is declared as a public class.
- - Context menu of the Acuminator has been updated: sort commands have been divided in the context menu, active sort indicator have been added, menu control templates have been redesigned to integrate the Visual Studio theme.
- - The following complementary information about DAC fields is displayed in a separate column: field type, whether the field is an identity field, whether the field is bound or unbound.
+ - The context menu of Acuminator has been updated: Sort commands have been divided in the context menu, active sort indicator has been added, and menu control templates have been redesigned to integrate the Visual Studio theme.
+ - The following complementary information about DAC fields is displayed in a separate column: the field type, an indicator of whether the field is an identity field, and an indicator of whether the field is bound or unbound.
  - It is now possible to sort the code map. The context menu now includes the following new commands: sort node's children by declaration, sort node's children by name, sort all node descendants by declaration, sort all node descendants by name.
 
 ### Fixed Bugs
 In this version of Acuminator, the following bugs have been fixed:
  - Acuminator failed on Microsoft Visual Studio 2017.
- - The Acuminator package was loaded in Visual Studio at random time including loading after the analysis was already executed.
+ - The Acuminator package was loaded in Visual Studio at a random time, including loading after the analysis was already executed.
 
 ## Acuminator 2.1.3
 Acuminator 2.1.3 includes the bug fixes and enhancements described in this section, as well as the features that have been implemented in previous versions.
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -9,7 +9,10 @@ In Acuminator 2.2, the following enhancements have been implemented:
  - The [PX1019](diagnostics/PX1019.md) diagnostic has been added to check for DAC fields of a non-string type with the `AutoNumberAttribute` attribute.
  - The [PX1020](diagnostics/PX1020.md) diagnostic has been added to check if the length of a string DAC field with the `AutoNumberAttribute` attribute is sufficient.
  - The [PX1022](diagnostics/PX1022.md) diagnostic has been added to check if a DAC extension or a graph extension is declared as a public class.
- - It is now possible to sort the code map. The context menu of the Acuminator Code Map now includes the following new commands: sort node's children by declaration, sort node's children by name, sort all node descendants by declaration, sort all node descendants by name. The context menu of Acuminator has been updated to include the sorting feature: The menu layout has been redesigned to better integrate with the Visual Studio style. Sort commands have been divided in the context menu, active sort indicator has been added, and menu control templates have been redesigned to integrate the Visual Studio theme.
+ - It is now possible to sort the code map. The context menu of the Acuminator Code Map now includes an active sort indicator and the following new commands: 
+   - Sort node's children by declaration order or by name
+   - Sort all node descendants by declaration order or by name
+   Also, the context menu has been redesigned to better integrate with the Visual Studio style.
  - The following complementary information about DAC fields is displayed in the Code Map in a separate column: the field type, an indicator of whether the field is an identity field, and an indicator of whether the field is bound or unbound.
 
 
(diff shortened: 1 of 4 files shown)


docs/ReleaseNotes.md
```diff
@@ -1,6 +1,23 @@
 # Acuminator Release Notes
 This document provides information about fixes, enhancements, and key features that are available in Acuminator.
 
+## Acuminator 2.2
+Acuminator 2.2 includes the bug fixes and enhancements described in this section, as well as the features that have been implemented in previous versions.
+
+### Enhancements
+In Acuminator 2.2, the following enhancements have been implemented:
+ - The [PX1019](diagnostics/PX1019.md) diagnostic has been added to check for DAC fields of non-strings types with the `AutoNumberAttribute` attribute.
```
SENya1990: I would say non-string type. Non-string**s** sounds like there are several string types

docs/ReleaseNotes.md
```diff
@@ -1,6 +1,23 @@
...
+ - The [PX1020](diagnostics/PX1020.md) diagnostic has been added to check if the length of a string DAC field with the `AutoNumberAttribute` attribute is sufficient.
+ - The [PX1022](diagnostics/PX1022.md) diagnostic has been added to check if a DAC or graph extension is declared as a public class.
```
SENya1990: I would write DAC extension or graph extension. Currently it sounds like the check is applied to either DAC or graph extension

docs/ReleaseNotes.md
```diff
@@ -1,6 +1,23 @@
...
+ - The context menu of Acuminator has been updated: Sort commands have been divided in the context menu, active sort indicator has been added, and menu control templates have been redesigned to integrate the Visual Studio theme.
```
SENya1990: There were no sort commands at all before this version, so, I guess it should be reworded. 
It's better to write that sort commands were added, and you can also add that context menu layout was redesigned to better integrate with Visual Studio style

docs/ReleaseNotes.md
```diff
@@ -1,6 +1,23 @@
...
+ - The following complementary information about DAC fields is displayed in a separate column: the field type, an indicator of whether the field is an identity field, and an indicator of whether the field is bound or unbound.
```
SENya1990: I would add "information  about DAC fields is displayed in Code Map in  a separate column"

docs/ReleaseNotes.md
SENya1990: I would also add a place where it changed - in a Code Map. Something like this:
The context menu of Acuminator Code Map has been updated

docs/ReleaseNotes.md
```diff
@@ -1,6 +1,23 @@
...
+ - It is now possible to sort the code map. The context menu now includes the following new commands: sort node's children by declaration, sort node's children by name, sort all node descendants by declaration, sort all node descendants by name.
```
SENya1990: Imho it's better merge this with the text above about context menu redesign. See the comment there

docs/ReleaseNotes.md
```diff
@@ -1,6 +1,23 @@
...
+
+### Fixed Bugs
+In this version of Acuminator, the following bugs have been fixed:
+ - Acuminator failed on Microsoft Visual Studio 2017.
+ - The Acuminator package was loaded in Visual Studio at a random time, including loading after the analysis was already executed.
```
SENya1990: I would rephrase this text because currently it describes not the fixed issue but the cause of the problem which is not relevant to the user. Imho, it's a bit more informative this way:
"The suppression file and Acuminator settings sometimes were loaded incorrectly due to package being loaded by Visual Studio at a random time"

docs/diagnostics/PX1019.md
```diff
@@ -0,0 +1,31 @@
...
+| Code   | Short Description                                                                                                | Type  | Code Fix    | 
+| ------ | ---------------------------------------------------------------------------------------------------------------- | ----- | ----------- | 
+| PX1019 | A DAC property field with the `AutoNumber` attribute must have the string type.                                  | Error | Available   |
+
+## Diagnostic Description
+You can use the `AutoNumber` attribute on a DAC field to enable auto-numbering for the field. The field must be of the string type.
+
+To fix the issue, you can do one of the following:
+ - Change the field type to string.
+ - Use the `AutoNumber` attribute on a different field that has a string type.
```
SENya1990: The description of code fix is missing. 
I suggest to describe it like in other diagnostics:
https://github.com/Acumatica/Acuminator/blob/7f20625dad642154001d19aa998a731cd1ee7964/docs/diagnostics/PX1011.md

docs/diagnostics/PX1019.md
```diff
@@ -0,0 +1,31 @@
...
+This document describes the PX1019 diagnostic.
+
+## Summary
+
+| Code   | Short Description                                                                                                | Type  | Code Fix    | 
+| ------ | ---------------------------------------------------------------------------------------------------------------- | ----- | ----------- | 
+| PX1019 | A DAC property field with the `AutoNumber` attribute must have the string type.                                  | Error | Available   |
+
+## Diagnostic Description
+You can use the `AutoNumber` attribute on a DAC field to enable auto-numbering for the field. The field must be of the string type.
```
SENya1990: People usually ask about the reason for the diagnostics, so it's good to provide one in the docs. We can write something like:
"The AutoNumber attribute expects a field of string type. Fields of other types are not supported"

docs/diagnostics/PX1020.md
```diff
@@ -0,0 +1,31 @@
...
+## Summary
+
+| Code   | Short Description                                                                                                | Type  | Code Fix    | 
+| ------ | ---------------------------------------------------------------------------------------------------------------- | ----- | ----------- | 
+| PX1020 | For the DAC property with the `AutoNumber` attribute, the string length must be greater than or equal to 15.     | Error | Available   |
+
+## Diagnostic Description
+You can use the `AutoNumber` attribute on a DAC field to enable auto-numbering for the field. The string must be of the string type and the field length you specify in the `PXDBString` attribute for the field must be at least 15 symbols.
+
+To fix the issue, you can make the string length sufficient for auto-numbering.
```
SENya1990: We need to describe code fix here.

docs/diagnostics/PX1020.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Summary
+
+| Code   | Short Description                                                                                                | Type  | Code Fix    | 
+| ------ | ---------------------------------------------------------------------------------------------------------------- | ----- | ----------- | 
+| PX1020 | For the DAC property with the `AutoNumber` attribute, the string length must be greater than or equal to 15.     | Error | Available   |
+
+## Diagnostic Description
+You can use the `AutoNumber` attribute on a DAC field to enable auto-numbering for the field. The string must be of the string type and the field length you specify in the `PXDBString` attribute for the field must be at least 15 symbols.
+
```
SENya1990: Here we need more details. First of all, the sufficient length is not a hard-coded number, 15. It is actually taken from `PX.Objects.CS.NumberingSequence` DAC's field `StartNbr`. For autonumbering to work correctly the length of the auto-numbered field can't be lesser than the length of a start  number of numbering sequence which is described by the `StartNbr` field. Currently, its length is 15, but its not hard-coded, the check actually obtains the value from the `NumberingSequence` DAC. 
We need to include this into the docs to explain users why the alert is displayed.

docs/diagnostics/PX1022.md
```diff
@@ -0,0 +1,37 @@
+# PX1022
+This document describes the PX1022 diagnostic.
+
+## Summary
+
+| Code   | Short Description                                                                                                | Type  | Code Fix    | 
+| ------ | ---------------------------------------------------------------------------------------------------------------- | ----- | ----------- | 
+| PX1022 | The non-public graph extensions are not supported.                                                               | Error | Available   |
```
SENya1990: the non public graph and **DAC** extensions are not supported

docs/diagnostics/PX1022.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Diagnostic Description
+You can declare a class that is a graph extension. The class must be public.
```
SENya1990: We need the explanation like:
"The non-public graph/DAC extensions are not loaded and wired up to graphs/DACs by platform" or
"The Acumatica Platform recognizes only public DAC/graph extensions"

docs/diagnostics/PX1022.md
```diff
@@ -0,0 +1,37 @@
...
+
+To fix the issue, you can make the graph extension public.
```
SENya1990: Code fix description is missing

docs/diagnostics/PX1020.md
```diff
@@ -8,10 +8,12 @@ This document describes the PX1020 diagnostic.
 | PX1020 | For the DAC property with the `AutoNumber` attribute, the string length must be greater than or equal to 15.     | Error | Available   |
 
 ## Diagnostic Description
-You can use the `AutoNumber` attribute on a DAC field to enable auto-numbering for the field. The string must be of the string type and the field length you specify in the `PXDBString` attribute for the field must be at least 15 symbols.
+You can use the `AutoNumber` attribute on a DAC field to enable auto-numbering for the field. For autonumbering to work correctly, the length of the auto-numbered field should not be lesser than the length of a start number of numbering sequence which is specified by the `StartNbr` field of the `PX.Objects.CS.NumberingSequence` DAC.
 
 To fix the issue, you can make the string length sufficient for auto-numbering.
 
+The code fix puts the sufficient length (specified by the `StartNbr` field of the `PX.Objects.CS.NumberingSequence` DAC) as the first parameter value for the `PXDBString` attribute.
```
SENya1990: I would rephrase it a little more.  It works also with PXString attribute and should work with attributes derived from PXString/PXDBString.  
It's not very good to rely explicitly on parameters order, so I would just say that the code fix replaces the length value passed to the string attribute's (PXString/PXDBString or derived from them)  constructor with the sufficient length. Which is taken from  `StartNbr` field of the `PX.Objects.CS.NumberingSequence` DAC

docs/ReleaseNotes.md
```diff
@@ -6,17 +6,17 @@ Acuminator 2.2 includes the bug fixes and enhancements described in this section
...
- - It is now possible to sort the code map. The context menu now includes the following new commands: sort node's children by declaration, sort node's children by name, sort all node descendants by declaration, sort all node descendants by name.
+ - The [PX1022](diagnostics/PX1022.md) diagnostic has been added to check if a DAC extension or a graph extension is declared as a public class.
+ - It is now possible to sort the code map. The context menu of the Acuminator Code Map now includes the following new commands: sort node's children by declaration, sort node's children by name, sort all node descendants by declaration, sort all node descendants by name. The context menu of Acuminator has been updated to include the sorting feature: The menu layout has been redesigned to better integrate with the Visual Studio style. Sort commands have been divided in the context menu, active sort indicator has been added, and menu control templates have been redesigned to integrate the Visual Studio theme.
```
SENya1990: Actually the words about sort commands being divided and menu control templates changed are implementation details, so I would not include them. I would change it to something like this:
```
It is now possible to sort the code map. The context menu of the Acuminator Code Map now includes an active sort indicator and the following new commands: 
- Sort node's children by declaration order / by name;
- Sort all node descendants by declaration order / by name;
Also context menu was redesigned to better integrate with the Visual Studio style.

```

docs/diagnostics/PX1022.md
```diff
@@ -5,13 +5,15 @@ This document describes the PX1022 diagnostic.
...
-| PX1022 | The non-public graph extensions are not supported.                                                               | Error | Available   |
+| PX1022 | The non-public graph and DAC extensions are not supported.                                                       | Error | Available   |
 
 ## Diagnostic Description
-You can declare a class that is a graph extension. The class must be public.
+You can declare a class that is a graph extension. The Acumatica Customization Platform recognizes only public graph or DAC extensions.
 
 To fix the issue, you can make the graph extension public.
 
+The code fix adds the `public` key word to the class declaration.
```
SENya1990: It replaces the current accessibility modifier with a public for the class declaration and also, in case of nested types, for all containing non-public type declarations.

docs/ReleaseNotes.md
```diff
@@ -9,7 +9,10 @@ In Acuminator 2.2, the following enhancements have been implemented:
...
  - The [PX1022](diagnostics/PX1022.md) diagnostic has been added to check if a DAC extension or a graph extension is declared as a public class.
- - It is now possible to sort the code map. The context menu of the Acuminator Code Map now includes the following new commands: sort node's children by declaration, sort node's children by name, sort all node descendants by declaration, sort all node descendants by name. The context menu of Acuminator has been updated to include the sorting feature: The menu layout has been redesigned to better integrate with the Visual Studio style. Sort commands have been divided in the context menu, active sort indicator has been added, and menu control templates have been redesigned to integrate the Visual Studio theme.
+ - It is now possible to sort the code map. The context menu of the Acuminator Code Map now includes an active sort indicator and the following new commands: 
+   - Sort node's children by declaration order or by name
+   - Sort all node descendants by declaration order or by name
```
SENya1990: This part merged with next sentence.
```suggestion
   - Sort all node descendants by declaration order or by name
   
```

PR #364, opened 2020-04-10
Feature/atr 440 dev

SENya1990:


src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/SemanticModel/PXGraph/Enums/GraphSemanticModelCreationOptions.cs
```diff
@@ -0,0 +1,38 @@
...
+		/// <summary>
+		/// Collect all possible semantic data during construction of the graph semantic model.
+		/// </summary>
+		CollectAll = CollectGeneralGraphInfo | CollectProcessingDelegates | InferImplicitModels
+	}
+
+
+	public static class GraphSemanticModelCreationOptionsUtils
+	{
+		public static bool HasOption(this GraphSemanticModelCreationOptions creationOptions, GraphSemanticModelCreationOptions flagToCheck) =>
```
dropsonic, line 35: This method can be replaced with `Enum.HasFlag`

SENya1990: Yes but it will cause boxing
https://stackoverflow.com/questions/11665279/why-enums-hasflag-method-need-boxing

SENya1990: Well in general I disagree - we should try to decrease a number of heap allocations .VS is slow enough by itself. However, in this case it indeed doesn't look like a hot path since code map is shared between analyzers and is created only once both for analysis, navigation and code map

SENya1990: fixed

PR #366, opened 2020-04-15
Feature/atr 551 dev

SENya1990:


src/Acuminator/Acuminator.Vsix/Resources/VSIXResource.resx
```diff
@@ -300,4 +300,28 @@
   <data name="PropertyHasAutoNumberingExtraInfoTooltip" xml:space="preserve">
     <value>Auto-numbering is enabled for this field.</value>
   </data>
+  <data name="DacKeyIconTooltip" xml:space="preserve">
```
SENya1990, line 303: @EAndrosova  please review the following strings for graph view extra info labels and tooltips

EAndrosova: I would suggest changing the string to the following:
_Autonumbering is enabled for this field._
(When a word is a noun, you don't need to use a hyphen)

SENya1990: @EAndrosova  done

PR #370, opened 2020-05-26
Bugfix/atr 494 dev

SENya1990:


src/Acuminator/Acuminator.Analyzers/StaticAnalysis/BqlParameterMismatch/Helpers/ParametersCounter.cs
```diff
@@ -52,24 +53,46 @@ public ParametersCounter(PXContext pxContext)
...
+			/// </summary>
+			/// <param name="typeSymbol">The type symbol.</param>
+			/// <param name="cancellationToken">(Optional) The cancellation token.</param>
+			/// <returns/>
+			public bool CountParametersInTypeSymbolForIdentifierNode(ITypeSymbol typeSymbol, CancellationToken cancellationToken = default)
+			{
+				if (!IsCountingValid || typeSymbol == null || IsCancelled(cancellationToken))
+					return false;
+
+				PXCodeType? codeType = typeSymbol.GetColoringTypeFromIdentifier();
```
dropsonic, line 83: Feels strange to see `GetColoringTypeFromIdentifier` in `Acuminator.Analyzers`. Maybe we should rename it to smth like "GetCodeTypeFromIdentifier"?

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/BqlParameterMismatch/Helpers/ParametersCounterSyntaxWalker.cs
```diff
@@ -54,12 +54,25 @@ public override void VisitGenericName(GenericNameSyntax genericNode)
...
+				if (!ParametersCounter.CountParametersInTypeSymbolForGenericNode(typeSymbol, _cancellationToken))
 					return;
 
 				if (!_cancellationToken.IsCancellationRequested)
 					base.VisitGenericName(genericNode);
 			}
+
+			public override void VisitIdentifierName(IdentifierNameSyntax identifierNode)
+			{
+				if (_cancellationToken.IsCancellationRequested)
```
dropsonic, line 66: Why not `ThrowIfCancellationRequested`? I may be wrong but it seems that the walker can finish its work with unexpected results instead of global cancellation of the whole process.

SENya1990: Yes, it's quite old diagnostic, we certainly can do ThrowIfCancellationRequested. I'll change it

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/BqlParameterMismatch/Helpers/ParametersCounter.cs
SENya1990: Yes, it was created for coloring but was conveniently reused for BQL analysis. It's a good idea to rename it. But I suggest something more related to BQL like
GetBqlPartTypeFromIdentifier

SENya1990: No, we can't make it related only to BQL because there are Action and Graph in enum.
I'll use GetCodeTypeFromIdentifier then

SENya1990: Done

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/BqlParameterMismatch/Helpers/ParametersCounterSyntaxWalker.cs
SENya1990: Done

PR #374, opened 2020-06-18
Feature/atr 564 dev primary and foreign keys check

SENya1990:
First part of the suit of checks for the PK/FK API. It is more focused on the PK checks.


src/Acuminator/Acuminator.Analyzers/DiagnosticsShortName.resx
```diff
@@ -339,4 +339,19 @@
   <data name="PX1016" xml:space="preserve">
     <value>ExtensionDoesNotDeclareIsActiveMethod</value>
   </data>
+  <data name="PX1033" xml:space="preserve">
```
SENya1990, line 342: @EAndrosova please check short names for the diagnostics.

src/Acuminator/Acuminator.Analyzers/Resources.resx
```diff
@@ -543,4 +543,28 @@
   <data name="PX1016GraphExtensionTitle" xml:space="preserve">
     <value>A graph extension must include a public static IsActive method with the bool return type. Extensions which are constantly active reduce performance. Suppress the error if you need the graph extension to be constantly active.</value>
   </data>
+  <data name="PX1033Title" xml:space="preserve">
```
SENya1990, line 546: @EAndrosova please check messages for the diagnostics and code fixes

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/Descriptors.cs
```diff
@@ -193,6 +193,26 @@ public static class Descriptors
...
 
+		public static DiagnosticDescriptor PX1033_MissingDacPrimaryKeyDeclaration { get; } =
+			Rule("PX1033", nameof(Resources.PX1033Title).GetLocalized(), Category.Default, DiagnosticSeverity.Info,
+				DiagnosticsShortName.PX1033);
+
+		public static DiagnosticDescriptor PX1034_MissingDacForeignKeyDeclaration { get; } =
+			Rule("PX1034", nameof(Resources.PX1034Title).GetLocalized(), Category.Default, DiagnosticSeverity.Info,
+				DiagnosticsShortName.PX1034);
+
+		public static DiagnosticDescriptor PX1035_MultiplePrimaryKeyDeclarationsInDac { get; } =
```
dropsonic, line 204: I think that PX1035 and PX1036 should be warnings. We don't want to force the developers to add PK/FK everywhere but in the case when they added it, PK/FK should be defined in the right way.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/Descriptors.cs
```diff
@@ -193,6 +193,26 @@ public static class Descriptors
...
+			Rule("PX1035", nameof(Resources.PX1035Title).GetLocalized(), Category.Default, DiagnosticSeverity.Info,
+				DiagnosticsShortName.PX1035);
+
+		public static DiagnosticDescriptor PX1036_WrongDacPrimaryKeyName { get; } =
```
dropsonic, line 208: Why do missing PK and FK diagnostics have different IDs but wrong name for PK and FK have the same one?

SENya1990: It's indeed something that can be discussed. Imho, the wrong key name diagnostic has more things in common for primary key/foreign key than the diagnostic for missing key declaration. Maybe, that' because the code fix for missing primary key is already complex enough and I don't want to include there logic for codefix for missing foreign key. And code fix for the wrong key name is rather simple, so it can be shared between primary and foreign keys.
If this code needs to be fixed than I would prefer to make a two separate check for wrong key name

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/Descriptors.cs
SENya1990: Ok. Should I write an email to them or ask in comments to the ATR-564 item?

dropsonic: Either way is fine; please emphasize the fact that this item is special since the main diagnostic is informational and intended to help all application teams with PK/FK implementation that is planned for Q3.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/Descriptors.cs
dropsonic: I think it's ok, I was just curious about the reason behind that.

PR #376, opened 2020-07-13
Feature/atr 564 dev part 2 fk code fix

SENya1990:


src/Acuminator/Acuminator.Analyzers/Resources.resx
```diff
@@ -567,4 +567,32 @@
   <data name="PX1035Fix" xml:space="preserve">
     <value>Remove other primary key declarations from the DAC</value>
   </data>
+  <data name="PX1034Fix" xml:space="preserve">
```
SENya1990, line 570: @EAndrosova please check code fix messages and foreign key example template. 
The template should look like this:
```csharp
public static class FK
{
  // Add all DAC foreign keys here. For referenced DACs with PK you can use the following template:
  // public class ReferencedDFK : ReferencedDAC.PK.ForeignKeyOf<SOOrder>.By<joinField1, joinFields, ...> { }
  //
  // For referenced DACs without PK class and with a single field primary key you can use the following template:
  // public class ReferencedDacFK : Field<joinField>.IsRelatedTo<ReferencedDAC.keyField>.AsSimpleKey.WithTablesOf<ReferencedDAC, SOOrder> { }
  //
  // For referenced DACs without PK class and with a composite primary key you can use the following template:
  /* public class ReferencedDacFK : CompositeKey<
                                                           Field<joinField1>.IsRelatedTo<ReferencedDAC.keyField1>,
                                                           Field<joinField2>.IsRelatedTo<ReferencedDAC.keyField2>,
                                                          ...
                                                          >.WithTablesOf<SOOrder, SOOrder> { } */
  //
  // You can start with the following properties:
  // OrderType,
  // BranchID,
  // CustomerID,
  // CustomerLocationID
}
```

PR #385, opened 2020-10-26
documentation fixes

SENya1990:


docs/diagnostics/PX1034.md
```diff
@@ -70,7 +70,7 @@ public class SOOrder : IBqlTable
 
 ## Related Articles
 
-[Foreign Keys and Nullable Columns](https://help.acumatica.com/(W(12))/Help?ScreenId=ShowWiki&pageid=8da9e9c6-ebbf-409a-b43d-a13d2081a62e)
```
dropsonic, line 73: What has been changed here? Line endings?

dropsonic (approved):

docs/diagnostics/PX1034.md
EAndrosova: Yes, I had to change them so that all links are displayed in a list

SENya1990: @dropsonic Yes, Github understand two whitespaces as a line break. Try look at the rich diff to see the difference

SENya1990: @EAndrosova actually I found that two spaces at the end work as well and the distance between lines is less. This imho makes the list look better

EAndrosova: I agree, it seems that only oxygen does not understand white space in markdown

EAndrosova (approved):

PR #391, opened 2021-01-13
Feature/atr 578 dev pxinternaluseonly

SENya1990:


docs/dev/CodingGuidelines/CodingGuidelines.md
```diff
@@ -345,4 +346,13 @@ public Task<(bool IsSuccess, TResult Result)> TryAwait<TResult>(this Task<TResul
 ```
 
 If you call it from Acuminator.Vsix then in Visual Studio 2017 the `MissingMethod` exception will be thrown. However, everything will be ok in Visual Studio 2019.
-Therefore, until we drop the support for Visual Studio 2017, do not declare public API containing value tuples. Either declare it as internal API or create custom structs for it.
\ No newline at end of file
+Therefore, until we drop the support for Visual Studio 2017, do not declare public API containing value tuples. Either declare it as internal API or create custom structs for it.
+
+### Debugging Hints
```
SENya1990, line 351: @EAndrosova please review addition to the coding guidelines

src/Acuminator/Acuminator.Analyzers/DiagnosticsShortName.resx
```diff
@@ -363,4 +363,7 @@
   <data name="PX1037" xml:space="preserve">
     <value>UnboundDacFieldInKeyDeclaration</value>
   </data>
+  <data name="PX1076" xml:space="preserve">
+    <value>CallToApiMarkedWithPXInternalUseOnly</value>
```
SENya1990, line 367: @EAndrosova please review diagnostic short name

src/Acuminator/Acuminator.Analyzers/Resources.resx
```diff
@@ -616,4 +616,7 @@
   <data name="PX1036FKFix" xml:space="preserve">
     <value>Place all foreign keys in a public static class named "FK"</value>
   </data>
+  <data name="PX1076Title" xml:space="preserve">
+    <value>Calls to Acumatica internal API marked with PXInternalUseOnlyAttribute are forbidden</value>
```
SENya1990, line 620: @EAndrosova  please review diagnostic warning message

README.md
```diff
@@ -73,6 +73,7 @@ You can navigate to every code element displayed in the Code Map by double click
 To build the solution, do the following:
 1. Create the _lib_ folder in the root folder. 
 2. Add _PX.Data.dll_, _PX.Data.BQL.Fluent.dll_, _PX.Common.dll_, _PX.BulkInsert.dll_, _PX.Objects.dll_, and _PX.DbServices.dll_ (from Acumatica ERP 2019 R1 or higher) to the _lib_ folder.
+   Starting from Acumatica ERP 2020R2 you also need to add _PX.Common.Std.dll_.
```
SENya1990, line 76: @EAndrosova please review changes in Readme file

src/Acuminator/Acuminator.Analyzers/Resources.resx
dropsonic: I think we need to elaborate further on this. This string is too strict (not polite enough) for external developers. I think it should be something like `This code calls Acumatica internal API marked with PXInternalUseOnlyAttribute that is not intended for public use`. @EAndrosova what do you think?

src/Acuminator/Acuminator.Analyzers/DiagnosticsShortName.resx
dropsonic: `PXInternalUseOnly` is, basically, a matter of implementation that might be changed later. I suggest "CallToInternalApi" or something like that

src/Samples/PX.Objects.HackathonDemo/PX.Objects.HackathonDemo/CallsToInternalAPI/InternalAPI.cs
```diff
@@ -0,0 +1,43 @@
...
+namespace PX.Objects.HackathonDemo.Graph.InternalApiUsage
+{
+    [PXInternalUseOnly]
+    public class InternalService 
+    {
+		public bool IsActive;
+
+		public bool SomeFlag { get; }
+
+        public void ProvideService()
```
dropsonic, line 14: A mix of whitespaces/tabs

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/CallsToInternalAPI/InternalApiCallsWalker.cs
```diff
@@ -0,0 +1,376 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Acuminator.Analyzers.StaticAnalysis.CallsToInternalAPI
+{
+	internal class InternalApiCallsWalker : CSharpSyntaxWalker
+	{
+		private static readonly ConcurrentDictionary<ISymbol, bool> _markedInternalApi = new ConcurrentDictionary<ISymbol, bool>();
```
dropsonic, line 27: If one solution is closed and another solution is opened, is this collection cleared?
What happens if somebody marked a type as `PXInternalUseOnly`, then the analysis was performed, and then `PXInternalUseOnly` was removed? Won't it lead to a symbol in this collection that has `true` as a value but in fact is no longer marked with `PXInternalUseOnly` in the source code?
What if the symbol was completely removed from the code? May it lead to a memory leak? Have you considered using `ConditionalWeakTable` instead?

src/Acuminator/Acuminator.Tests/Verification/VerificationHelper.cs
```diff
@@ -28,9 +30,23 @@ public static class VerificationHelper
 		private static readonly MetadataReference PXDataReference = MetadataReference.CreateFromFile(typeof(PXGraph).Assembly.Location);
 		private static readonly MetadataReference FluentBqlReference = MetadataReference.CreateFromFile(typeof(FbqlCommand).Assembly.Location);
 		private static readonly MetadataReference PXCommonReference = MetadataReference.CreateFromFile(typeof(PX.Common.PXContext).Assembly.Location);
+		private static readonly MetadataReference PXCommonStdReference = MetadataReference.CreateFromFile(typeof(PX.Common.PXInternalUseOnlyAttribute).Assembly.Location);
 		private static readonly MetadataReference PXObjectsReference =
 			MetadataReference.CreateFromFile(typeof(PX.Objects.GL.PeriodIDAttribute).Assembly.Location);
 
+		private static readonly MetadataReference[] _metadataReferences =
```
dropsonic, line 37: Since you moved all the references to an array, do we still need them as separate `static readonly` properties? I think the whole array can consist of `MetadataReference.CreateFromFile(...)` and similar calls to increase readability - right now a developer need to follow the links 2 times (`AddMetadataReferences` -> `_metadataReferences` -> a particular private field)  to get the idea of what references are actually added to the project.

src/Samples/PX.Objects.HackathonDemo/PX.Objects.HackathonDemo/CallsToInternalAPI/SOShipmentExt.cs
```diff
@@ -0,0 +1,36 @@
...
+			}
+
+            return filesNbr;
+        }
+
+        public void InvalidateCache()
+        {
+			if (serviceProvider.AccessChecker != null && serviceProvider.AccessChecker.IsActive && serviceProvider.AccessChecker.ShouldCheck)
+			{
+                serviceProvider.AccessChecker.CheckAccess();
```
dropsonic, line 25: A mix of whitespaces and tabs.

dropsonic (changes requested):

src/Acuminator/Acuminator.Analyzers/Resources.resx
SENya1990: Maybe we should add something like "we give no guarantee that it won't be changed in the future" or "is subject to changes in the future"?

dropsonic: I think the details about possible consequences can be explained in the diagnostic's description. It is too long and cumbersome to fit it to the title itself.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/CallsToInternalAPI/InternalApiCallsWalker.cs
SENya1990: I was more concerned about parallel access when I wrote this cache and thought about internal API as of something static - that this is a check in ISV mode for external devs who don't have control over internal API and have no reason to use the `PXInternalUseOnly`. Therefore, since the external developer can't change the internal API everything looks not that bad. The cache can be considered as a register for internal/non internal API. 
There is a potential memory leak in the cases you mentioned - when we switch solutions, or make changes to the internal API (imho is a corner case). For the former case I think the `ConditionalWeakTable` is the solution and it is worth to change the code to use it especially since it is thread safe. 

As for the latter case - there is a problem. If we need to consider that the external dev can change the internal API then there is no possibility to cache symbols. We don't explicitly control when the symbol will be deleted from the `ConditionalWeakTable` nor can we somehow know beforehand that internal API was changed and update the cache. It is possible that the developer deleted `PXInternalUseOnly` attribute or the whole API with this attribute but during the analysis it is still there because GC hasn't happened yet.
Maximum we can do in such case is to:
1. Stop caching non internal API which is a performance hit but not a very big one I think
2. Cache only internal API from Acumatica. Check for namespaces for example. It isn't very convenient and we have to either keep a list of namespaces updated or rely on the **PX.** prefix and hope that no one else will use it. There will be small additional overhead of checking namespaces so it is possible that in such case it is best to not cache at all.
Or we can mark `PXInternalUseOnly` attribute itself with `PXInternalUseOnly`  to set in stone that only we control what is an internal API and what isn't.

src/Acuminator/Acuminator.Tests/Verification/VerificationHelper.cs
SENya1990: The logic is a bit more complex - there is a `BuildAssemblyFromSources` method which creates a separate `MetadataReference` array with almost all of this references present (excluding `FluentBqlReference`). However, I think it's possible to change the code and use the same metadata references everywhere. So I'll fix it

src/Samples/PX.Objects.HackathonDemo/PX.Objects.HackathonDemo/CallsToInternalAPI/InternalAPI.cs
SENya1990: fixed

src/Samples/PX.Objects.HackathonDemo/PX.Objects.HackathonDemo/CallsToInternalAPI/SOShipmentExt.cs
SENya1990: fixed

src/Acuminator/Acuminator.Tests/Verification/VerificationHelper.cs
SENya1990: I have tried to make an array of `MetadataReference.CreateFromFile` and actually I don't think that it will be more helpful for the developer. When I look at the 
```csharp
MetadataReference.CreateFromFile(typeof(PX.Common.PXInternalUseOnlyAttribute).Assembly.Location)
```
I don't know what assembly I'll get. For the majority of the symbols I don't know which assemblies contain which types and imho other devs also don't remember this. If I try to guess the assembly for `PX.Common.PXInternalUseOnlyAttribute` I would say that it is located in 'PX.Common' due to namespace name. But it isn't this way now. I need to make the same "Go to" to the `PX.Common.PXInternalUseOnlyAttribute` symbol's metadata to see its assembly. Or I have to write assembly name in the comment to keep the info about the assembly in the same location. Neither is better than existing solution in my opinion when we already have a meaningful names for the assembly fields

src/Acuminator/Acuminator.Analyzers/Resources.resx
EAndrosova: @dropsonic  I like your suggestion, I will replace the string as part of the review

src/Acuminator/Acuminator.Analyzers/DiagnosticsShortName.resx
SENya1990: fixed

EAndrosova (approved):

src/Acuminator/Acuminator.Tests/Verification/VerificationHelper.cs
dropsonic: I also dislike getting an assembly reference from a specific type. It seems more logical to use relative paths since it is more readable.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/CallsToInternalAPI/InternalApiCallsWalker.cs
dropsonic: As we discussed previously, our own developer may change `PXInternalUseOnly` marking on some type when working with `PureWithNetTools.sln` or open the same solution from another branch with a different set of `PXInternalUseOnly`-marked types so the cache cannot be global.
I think that, first of all, we need to make sure that there is a performance issue before fixing it in a complex way.

dropsonic:
@SENya1990 do you correctly handle a case when a class is marked with `PXInternalUseOnly` but its members don't?
An example:
image

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/CallsToInternalAPI/InternalApiCallsWalker.cs
SENya1990: I have changed the used cache to `ConditionalWeakTable`. Now even when there are changes to the document the cache is invalidated and old entries are collected. It also works this way when solution changes. The reload of solution also happens when you switch git branches. I have also checked that the diagnostic works with the manual changes of the code like addition of `PXInternalUseOnly` attribute or its removal. We also don't have to worry about multiple instances of VS since each of them has its own memory and therefore cache. Even if the out-of-process feature is turned on there still will be two different code analysis processes for two different Visual Studios
As for the performance issues - it is hard to measure something precisely with diagnostics and it is a separate big amount of work. In my opinion this particular diagnostic should use global caching because:
1. It checks the whole syntax tree, not a particular symbol or node. It works for all types, not just types specific to Acumatica.
2. It works on IdentifierName nodes - the most frequently encountered node in the syntax tree. It also works on other common nodes - GenericName, MemberAccess. By work I mean it makes the most expensive checks here.
3. The checks are not that cheap either. To correctly deduce that the type belongs to the internal API we need to check its base type + its containing type. For each of this two types we also need to check their base type and containing type and etc. As you can see there is an exponential growth of types needed to be checked especially if the type doesn't belong to the internal API (which is the most common case) and we have to check all possible combination of these types. In practice their number is not that big but it still can sum up. For each checked access of a type member we also have to do the same check for its containing type. So the number of checks really grows

src/Acuminator/Acuminator.Tests/Verification/VerificationHelper.cs
SENya1990: But getting the reference from the type's assembly location is more stable - you can't get out of sync with file location. If you transfer the assembly files to some other place you have to remember that you must update the strings with paths to assemblies here. Moreover you can do this only for the Acumatica assemblies. It shouldn't be done with corlib assembly or System.Core or assemblies from nuget packages. So it won't be very consistent when about a half of assemblies is created with a string and a half with type's assembly location.

SENya1990:
@dropsonic Yes the check supports this case and it is even tested. If you look in the test sources for the diagnostic there is the InternalAPI.cs file where there is the following code:
изображение
The file GraphUsingInternalApi.cs located next to it contains calls to the non internal properties of the internal type:
изображение

dropsonic (approved):

PR #395, opened 2021-05-23
Feature/atr 612 dev rn for 2 3 2

EAndrosova:
Please review release notes for Acuminator 2.3.2 and the description for the PX1076 diagnostic.

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -1,6 +1,22 @@
 # Acuminator Release Notes
 This document provides information about fixes, enhancements, and key features that are available in Acuminator.
 
+## Acuminator 2.3.2
+Acuminator 2.3.2 includes the bug fixes and enhancements described in this section, as well as the features that have been implemented in previous versions.
+
+### Enhancements
+In Acuminator 2.3.2, the following enhancements have been implemented:
+ - The [PX1076](diagnostics/PX1076.md) diagnostic has been added to check for and warn a developer that the PXInternalUseOnly attribute is used in the customization code.
+ - The documentation for the [PX1046](diagnostics/PX1046.md) diagnostic was extended, the reasoning for the diagnostic is now provided.
+ 
+### Fixed Bugs
+In this version of Acuminator, the following bugs have been fixed:
+ - Due to the changes in the theme updataing mechanism, an error was thrown by Visual Studio in the Acuminator colorizer, and the Acuminator coloring became corrupted.
+ - Due to the changes in the lastest Visual Studio updates, Acuminator custom settings were not enabled, and the Acuminator analyzers used default values.
+ - In the [PX1036](diagnostics/PX1036.md) diagnostic, foreign keys for different DACs that use the same set of DAC fields were incorrectly considered as duplicates.
+ - Due to the changes in the latest Visual Studio updates, switching between themes was impossible. 
+
+
 ## Acuminator 2.3.1
 Acuminator 2.3.1 includes the bug fixes and enhancements described in this section, as well as the features that have been implemented in previous versions.
 
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -11,8 +11,8 @@ In Acuminator 2.3.2, the following enhancements have been implemented:
  
 ### Fixed Bugs
 In this version of Acuminator, the following bugs have been fixed:
- - Due to the changes in the theme updataing mechanism, an error was thrown by Visual Studio in the Acuminator colorizer, and the Acuminator coloring became corrupted.
- - Due to the changes in the lastest Visual Studio updates, Acuminator custom settings were not enabled, and the Acuminator analyzers used default values.
+ - Due to the changes in the theme updating mechanism, an error was thrown by Visual Studio in the Acuminator colorizer, and the Acuminator coloring became corrupted.
+ - Due to the changes in the latest Visual Studio updates, Acuminator custom settings were not enabled, and the Acuminator analyzers used default values.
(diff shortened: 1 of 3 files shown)


docs/ReleaseNotes.md
```diff
@@ -1,6 +1,22 @@
 # Acuminator Release Notes
 This document provides information about fixes, enhancements, and key features that are available in Acuminator.
 
+## Acuminator 2.3.2
+Acuminator 2.3.2 includes the bug fixes and enhancements described in this section, as well as the features that have been implemented in previous versions.
+
+### Enhancements
+In Acuminator 2.3.2, the following enhancements have been implemented:
+ - The [PX1076](diagnostics/PX1076.md) diagnostic has been added to check for and warn a developer that the PXInternalUseOnly attribute is used in the customization code.
```
SENya1990, line 9: This diagnostic does not check the `PXInternalUseOnly` attribute usage. 

Some Acumatica APIs are effectively internal but can't have `internal` access modifier because they are accessed from different projects. Such APIs has to be `public` from the C# point of view. This means that partner developers can access them easily without resorting to complex means such as .Net Reflection. However, we don't want external developers to use these internal APIs. We mark such APIs with `PXInternalUseOnly` attribute to emphasize that they are internal to Acumatica and should not be used in external solutions - customizations, integrations, etc. We won't maintain such internal APIs - they can be changed or deleted in any minor upgrade.  
The downside of the attribute approach is that the developer needs to check the declaration of any used Acumatica API for the presence of `PXInternalUseOnly`attribute. This is tiresome and is frequently forgotten leading to a situation when some external solutions rely on internal Acumatica APIs. The PX1076 diagnostic addresses this flaw by displaying a warning to external developers for all used Acumatica APIs marked with `PXInternalUseOnly`attribute. 

Note that it has nothing to do with the usage of `PXInternalUseOnly`attribute itself in the code of customizations or other external solutions. It is not recommended to mark new code in external solutions with this attribute but it is ok to check in the customization code if this attribute is defined on some class or property

docs/ReleaseNotes.md
```diff
@@ -1,6 +1,22 @@
...
+ - The documentation for the [PX1046](diagnostics/PX1046.md) diagnostic was extended, the reasoning for the diagnostic is now provided.
+ 
+### Fixed Bugs
+In this version of Acuminator, the following bugs have been fixed:
+ - Due to the changes in the theme updating mechanism, an error was thrown by Visual Studio in the Acuminator colorizer, and the Acuminator coloring became corrupted.
+ - Due to the changes in the latest Visual Studio updates, Acuminator custom settings were not enabled, and the Acuminator analyzers used default values.
```
SENya1990, line 15: I think we should specify Visual Studio version here. Also the root cause of the issue should be specified for this bug. I would say it like this:
`Acuminator custom code analysis settings specified by the user did not work with a new 64-bit out of process static code analysis introduced by Visual Studio 2019. When out of process feature was enabled only default settings were used.`

docs/ReleaseNotes.md
```diff
@@ -1,6 +1,22 @@
...
+Acuminator 2.3.2 includes the bug fixes and enhancements described in this section, as well as the features that have been implemented in previous versions.
+
+### Enhancements
+In Acuminator 2.3.2, the following enhancements have been implemented:
+ - The [PX1076](diagnostics/PX1076.md) diagnostic has been added to check for and warn a developer that the PXInternalUseOnly attribute is used in the customization code.
+ - The documentation for the [PX1046](diagnostics/PX1046.md) diagnostic was extended, the reasoning for the diagnostic is now provided.
+ 
+### Fixed Bugs
+In this version of Acuminator, the following bugs have been fixed:
+ - Due to the changes in the theme updating mechanism, an error was thrown by Visual Studio in the Acuminator colorizer, and the Acuminator coloring became corrupted.
```
SENya1990, line 14: This is an internal bug which imho should be excluded entirely or a trimmed to a very short description:
`Improved stability of Acuminator syntax highlighting`

docs/ReleaseNotes.md
```diff
@@ -1,6 +1,22 @@
...
+ - Due to the changes in the latest Visual Studio updates, Acuminator custom settings were not enabled, and the Acuminator analyzers used default values.
+ - In the [PX1036](diagnostics/PX1036.md) diagnostic, foreign keys for different DACs that use the same set of DAC fields were incorrectly considered as duplicates.
+ - Due to the changes in the latest Visual Studio updates, switching between themes was impossible.
```
SENya1990, line 17: I think the VS version should be specified:
`The change of a current Visual Studio theme didn't work correctly for Acuminator coloring in Visual Studio 2019`

SENya1990 (changes requested):

docs/ReleaseNotes.md
```diff
@@ -1,6 +1,22 @@
 # Acuminator Release Notes
 This document provides information about fixes, enhancements, and key features that are available in Acuminator.
 
+## Acuminator 2.3.2
+Acuminator 2.3.2 includes the bug fixes and enhancements described in this section, as well as the features that have been implemented in previous versions.
+
+### Enhancements
+In Acuminator 2.3.2, the following enhancements have been implemented:
+ - The [PX1076](diagnostics/PX1076.md) diagnostic has been added to check for and warn a developer that that their code uses Acumatica APIs marked with the PXInternalUseOnly attribute.
```
dropsonic, line 9: Excessive `that` word: `warn a developer that that their code ...`

dropsonic (changes requested):

dropsonic (approved):

dropsonic:
@SENya1990 if you're ok with these changes, let's merge

SENya1990 (approved):

PR #403, opened 2021-07-14
Resolved format error when VS Preview build is being used.

ShadowFiendZX:
Resolved format error when VS Preview build is being used.

### DIFF
--- a/src/Acuminator/Acuminator.Vsix/Utils/Version/VSVersionProvider.cs
+++ b/src/Acuminator/Acuminator.Vsix/Utils/Version/VSVersionProvider.cs
@@ -2,10 +2,7 @@
 
 using System;
 using System.IO;
-using System.Collections.Generic;
 using System.Diagnostics;
-using System.Linq;
-using System.Text;
 using System.Threading.Tasks;
 
 using Microsoft.VisualStudio;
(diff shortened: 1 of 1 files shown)


src/Acuminator/Acuminator.Vsix/Utils/Version/VSVersionProvider.cs
```diff
@@ -36,27 +36,68 @@ public static async Task<VSVersion> GetVersionAsync(IAsyncServiceProvider servic
...
+            {
+                /*
+                 * For Preview builds Microsoft returns something like "16.11.0 Preview 3.0".
+                 * The Version constructor gives a string format error because it doesn't expect something like this.
+                 * Looking at their source code shows that they accept "major.minor.build.revision".
+                 * This string Replace returns "16.11.0.3.0".
+                 */
+                versionString = versionString.Replace(" Preview ", ".");
+
+                StringBuilder stringBuilder = new StringBuilder();
```
SENya1990, line 74: This looks too complex, I think it can be simplified. I had in mind something like this:
```
public string Parse(string version)
. {
.     const string previewKeyWord = "Preview";
.     int previewIndex = version.IndexOf(previewKeyWord);
. 
.     if (previewIndex < 0)
.         return version;
. 
.     int lastSegmentBeforePreviewIndex = version.LastIndexOf('.', previewIndex);
. 
.     if (lastSegmentBeforePreviewIndex < 0)
.         return version;
. 
.     string firstPart = version.Substring(0, lastSegmentBeforePreviewIndex + 1);
.     string secondPart = version.Substring(previewIndex + previewKeyWord.Length + 1);
.     return firstPart + secondPart;
. }
```

src/Acuminator/Acuminator.Vsix/Utils/Version/VSVersionProvider.cs
```diff
@@ -36,27 +36,68 @@ public static async Task<VSVersion> GetVersionAsync(IAsyncServiceProvider servic
...
             return new VSVersion(unknownVersion);
         }
 
+        private static string ParseVersion(object? versionObj)
+        {
+            string versionString = versionObj?.ToString() ?? "";
+
+            if (versionString.Contains("Preview"))
+            {
+                /*
```
SENya1990, line 66: Good comment. But it would be even more helpful if it was done as XML comment on the method, extra details may go into the <remark> section.

src/Acuminator/Acuminator.Vsix/Utils/Version/VSVersionProvider.cs
ShadowFiendZX: I'd had something like that initially but if you return 16.11.0.3.0 the Version constructor will give an error because it has too many segments. I suppose you could try to get a further substring out of secondPart, but I'd assumed that it'd be faster to search through the string one time in a for loop and break out of it when we've satisfied the requirements.

SENya1990: Oops, this will make the string "16.11.3.0". If we wish to make it "16.11.0.3" it will be a bit different
```c#
public string Parse(string version)
. {
.     const string previewKeyWord = "Preview";
.     int previewIndex = version.IndexOf(previewKeyWord);
. 
.     if (previewIndex < 0)
.         return version;
. 
.     string firstPart = version.Substring(0, previewIndex - 1);
.     int indexOfDotInPreviewVersion = version.IndexOf('.', previewIndex);
.     int secondPartStart = previewIndex + "Preview".Length + 1;
.     string secondPart = indexOfDotInPreviewVersion < 0
. 		? version.Substring(secondPartStart)
. 	    : version.Substring(secondPartStart, indexOfDotInPreviewVersion - secondPartStart);
.     return $"{firstPart}.{secondPart}";
. }
```

SENya1990 (approved):

dropsonic (approved):

dropsonic:
@ShadowFiendZX thank you!

PR #412, opened 2021-11-30
update packages for VS2022 support

shawnburt:
Update manifest to have 64bit 2022 listings
Update VSPackage
use dummy signing key


SENya1990:
The PR should not be merged until an explicit notification

src/Acuminator/Acuminator.Analyzers/Acuminator.Analyzers.csproj
```diff
@@ -8,19 +8,19 @@
     <Copyright>Copyright © 2017-2019 Acumatica Ltd.</Copyright>
     <LangVersion>8.0</LangVersion>
     <SignAssembly>true</SignAssembly>
-    <AssemblyOriginatorKeyFile>..\..\key.snk</AssemblyOriginatorKeyFile>
```
SENya1990, line 11: The type of the key should not be changed, since our build infrastructure and all other participants use .snk key

src/Acuminator/Acuminator.Analyzers/Acuminator.Analyzers.csproj
```diff
@@ -8,19 +8,19 @@
...
+    <AssemblyOriginatorKeyFile>..\key.pfx</AssemblyOriginatorKeyFile>
     <Company>Acumatica</Company>
     <Description>The static analyzers for Acumatica</Description>
     <Authors>Acumatica</Authors>
     <NeutralLanguage>en</NeutralLanguage>
     <RestorePackagesPath>../../packages</RestorePackagesPath>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis" Version="2.10.0" />
```
SENya1990, line 19: Using newer Roslyn nugets will drop the support for previous Visual Studios

SENya1990:
@shawnburt There is a PR with support for three VS versions:
https://github.com/Acumatica/Acuminator/pull/415

Unfortunately I didn't manage to use the changes from this PR there (except the addition of .editorconfig file). The approaches are too different. Here the support is added only for VS 2022. In the other PR the whole approach to dependencies on VS SDK is reworked, all individual packages are replaced with a single aggregate package. It is quite important as the PR excludes the  dependency on VS SDK package from run-time thus allowing to use different SDK packages that are shipped with different versions of Visual Studio.

Therefore I will close this PR. We will use the other one because it adds support for all versions of VS in one package. 
Still many people used this beta version and it works really well with VS 2022.  @shawnburt thank you very much for it! This gave us some time and allowed developers working with Acumatica to migrate to VS 2022

dropsonic:
@shawnburt I'd also like to thank you for taking the initiative and helping us with the migration! Thanks a lot!

PR #417, opened 2022-02-03
Feature/atr 432 dev is active code map

SENya1990:
Quick overview:

- Added IsActiveMethodInfo DTO to the core utils and integrated it into semantic models and analyzers
- Added new grouping node for graph "Initialize and Activate" category
- Added new graph member node for IsActive method which is displayed in the new graph "Initialize and Activate" category
- Extended visitors to support new nodes
- Integrated new nodes with tree builder
- Added two icons from VS icons set (will discuss with Valery Vydrin after his return from vacation)
- Completely reworked tooltips. Added mechanism for trimming excessive text in node tooltip. 
- Separated tooltips design from code map specific elements.
- Added caching of tooltips for attribute nodes
- Improved algorithm for removal of excessive indentation in tooltips
- Refactoring


src/Acuminator/Acuminator.Vsix/Tool Windows/CodeMap/Model/GraphMemberType.cs
```diff
@@ -30,7 +31,8 @@ internal static class GraphMemberTypeUtils
 			{ GraphMemberType.RowEvent, "Row Events" },
 			{ GraphMemberType.FieldEvent, "Field Events" },
 			{ GraphMemberType.NestedDAC, "Nested DACs" },
-			{ GraphMemberType.NestedGraph, "Nested Graphs" }	
+			{ GraphMemberType.NestedGraph, "Nested Graphs" },
+			{ GraphMemberType.SpecialMember, "Special Members" }
```
SENya1990, line 35: @dropsonic I'm not particularly sure that name "Special Members" is the best. It is planned as a category that unifies different special things related to graphs and DACs such as:

-  IsActive methods for extensions
- Graph constructors. We can make a separate category for them, but in 99.9% of cases (in correctly written code) we don't expect more than one instance parameterless constructor.
- Initialize methods for extensions
- Other specific things we may add later.

Maybe there is a better name?

dropsonic: We also have `IGraphWithInitialization.Initialize` in graphs.
Maybe something like "Initialization"? `IsActive` is also partially related to that.

I'd suggest asking @EAndrosova, maybe she has some ideas.

SENya1990: I forgot to mention `Persist` overrides. We would like to display them too. A general purpose group for specific members that can be encountered no more than once is imho a more universal solution. 
Also interested in @EAndrosova opinion.

dropsonic: What about `Load`/`Unload` overrides? In this case `Special Members` looks good. However, we still can extract graph constructors, `IsActive` and two `Initialize` methods into a separate group called `Initialization&Activation`, for example.

SENya1990: Load/Unload overrides imho are rare, so I'm not sure we should support them. 
However, I like the idea of having a separate group for initialization, we have many graph members related to it. 
I think it should be adopted. 

I'm not sure we will need Special Members category in this case, we can make another category for overrides and put Persist overrides there as well as overrides for ExecuteSelect and other graph methods. Let's not think about it now.

dropsonic: I really like the idea of having a separate "overrides" group.

SENya1990: Then I will make changes. @dropsonic , @EAndrosova  what do you think about these names:
- "Initialization & Activation" for graph members related to graph/DAC/extension initialization and activation
- "Base Overrides" for overrides of base class members for graph/graph extension. I don't want people to confuse it with "PXOverrides" category

SENya1990: Here is how "Initialization & Activation" will look. Notice a tooltip with method implementation
image

SENya1990:
Updated PR with following changes:
- replaced "Special Members" category with "Initialization & Activation", changed category icon
- removed caching mechanism because WPF already caches tooltips internally
- reworked tooltips into a separated mechanism independent from Code Map
- added mechanism to trim excessive text, integrated it with tooltips

src/Acuminator/Acuminator.Vsix/Tool Windows/CodeMap/Model/GraphMemberType.cs
dropsonic: > Here is how "Initialization & Activation" will look. Notice a tooltip with method implementation image

Looks awesome!

Btw, it seems that the tooltip uses 8 spaces for a tab instead of 4 spaces.

SENya1990: Yes, there is some extra indent. The thing is you have to remove excessive indentation from the tooltip string. There is  one indentation level for code inside namespace, one for type member inside a type and more levels if we are inside nested classes. Also sometimes formatting can be broken. It's not so easy to remove indent in a general way. I'll try to improve the removal but I'll do it in scope of other PR since I have already started to add "Initialization & Activation" for DACs

SENya1990: Actually indent removal algorithm works correctly, it's just a peculiarity of WPF rendering which renders tab indent much bigger than indentation that consists of whitespaces.

EAndrosova: After I read first comments where you listed all possible methods, I wanted to suggest "Utility Methods" or "System Methods" category.
But separating this category into 2 different is better. I agree with "Initialization & Activation" and "Base Overrides" titles.

EAndrosova (approved):

PR #420, opened 2022-02-07
Feature/atr 416 dev graph constructor

SENya1990:
Merge after ATR-650 (https://github.com/Acumatica/Acuminator/pull/419)

Changes:
- added display of graph instance and static constructors


dropsonic (approved):

SENya1990:
There is no rule against them and there are some places in code base that have them. They can be used for initialization of some static graph data.  Since we already collect info about static constructors in the graph semantic model I think it won't hurt to show it.

PR #444, opened 2022-04-27
Feature/atr 588 dev check is active for graph creation

SENya1990:
Overview:
- introduced new PX1056 diagnostic that forbids graph creation in `IsActive` of graph and DAC extensions and `IsActiveForGraph<TGraph>` methods of graph extensions. The check detects both direct graph constructor calls and `PXGraph.CreateInstance` calls to prevent expressions like `new PXGraph()` 
- made existing PX1057 and PX1084 checks a bit more strict by forbidding graph creation via direct constructor calls (`new PXGraph()` )
- enhanced graph semantic model with `IsActiveForGraph` method. Will be useful for display in Code Map.
- increased version to 3.1.0
-  refactoring


src/Acuminator/Acuminator.Analyzers/DiagnosticsShortName.resx
```diff
@@ -366,4 +366,10 @@
   <data name="PX1076" xml:space="preserve">
     <value>CallToInternalApi</value>
   </data>
+  <data name="PX1056IsActiveMethod" xml:space="preserve">
```
SENya1990, line 369: @EAndrosova please review diagnostic short names

src/Acuminator/Acuminator.Analyzers/Resources.resx
```diff
@@ -619,4 +619,10 @@
   <data name="PX1076Title" xml:space="preserve">
     <value>This code calls Acumatica internal API marked with PXInternalUseOnlyAttribute which is not intended for public use</value>
   </data>
+  <data name="PX1056TitleIsActive" xml:space="preserve">
```
SENya1990, line 622: @EAndrosova please review diagnostic messages

dropsonic (approved):

src/Acuminator/Acuminator.Analyzers/Resources.resx
SENya1990: @EAndrosova the new diagnostic for review is actually PX1056:
`A PXGraph instance cannot be initialized inside the IsActive method`
and 
`A PXGraph instance cannot be initialized inside the IsActiveForGraph&lt;TGraph&gt; method`

The one you mentioned is for already released diagnostic so it was reviewed. The main purpose of that diagnostic is to prevent external developers from using our internal API, reserved by Acumatica for its own use. It targets external developers only. I think it's worth to clarify for them which company reserves the API since in general case there could be internal APIs from other companies like Microsoft. Plus it is already released. Still we can change it if needed.

docs/diagnostics/PX1056.md
```diff
@@ -0,0 +1,43 @@
...
+This document describes the PX1056 diagnostic.
+
+## Summary
+
+| Code   | Short Description                          | Type  | Code Fix  | 
+| ------ | ------------------------------------------ | ----- | --------- | 
+| PX1056 | A PXGraph instance cannot be initialized inside the IsActive or IsActiveForGraph<TGraph> method. | Error | Unavailable | 
+
+## Diagnostic Description
+You cannot create a graph instance in the `IsActive` or `IsActiveForGraph<TGraph>` method because it can lead to a deadlock.
```
SENya1990, line 11: @EAndrosova may be we should mention that `IsActive` methods are declared in extensions? Something like this:
```suggestion
You cannot create a graph instance in the `IsActive` method of a graph or DAC extension and in the `IsActiveForGraph<TGraph>` method of a graph extension because it can lead to a deadlock. 
```
Although I'm not sure. What do you think?

EAndrosova (approved):

PR #445, opened 2022-05-23
Feature/atr 15 dev

SENya1990:
Code Snippets for Acumatica.
They also will be presented at Virtual DevCon

Note that in general Code Snippets are buggy in Visual Studio especially when distributed with VSIX. 
I guess there is little we can do with it.


SENya1990:
@EAndrosova  please review changes in the readme file and descriptions in code snippets

dropsonic:
@SENya1990 do we need this PR since now we have [Acumatica/CodeSnippets](https://github.com/Acumatica/CodeSnippets)?

SENya1990:
@dropsonic  it would be good to share code snippets with Acuminator. The visibility of  [Acumatica/CodeSnippets](https://github.com/Acumatica/CodeSnippets) is limited - you need to be told about it from somewhere. Acuminator is much better known, it is recommended in our dev trainings. You also need to install the snippets manually correctly. It's not very convenient.  
Frankly, I believe that more people will used snippets from Acuminator than from a separate repository mentioned on DevCon. 
This is the main reason.

One of the reasons to create a separate repository for code snippets was bad VS support for snippets distribution with VSIX packages. If we create a custom snippet installation mechanism in Acuminator this problem will be solved. It can be done somewhere in the future since it's not an urgent task. 
Still, it is not that hard to do:
- Add snippet files as resources to assembly
- On package initialization check some predefined folder and create it if needed. 
For example .\Documents\Visual Studio {VSVersion}\Acumatica Code Snippets\
- Keep a small file with settings like version of Acuminator that installed the snippets here before. If current package version is higher then reinstall the snippets. 
So in the future the distribution with Acuminator will reach more people and it can be more stable as well.

At the same time I believe it is useful to have snippets in a separate place too. People can suggest snippets there and it is not tied to Acuminator releases. It also shouldn't take many resources for synchronizing snippets between Acuminator and Acumatica/CodeSnippets. We can always take snippets from Acumatica/CodeSnippets only before the release of new version of Acuminator. We can even configure GitHub actions although I  don't think we need it.

SENya1990:
AFAIK @EAndrosova  should have pushed the changes here first. It's strange

dropsonic (approved):

PR #452, opened 2022-08-22
Bugfix/atr 670 dev fix alert on capture of non graphs

SENya1990:
There is a lot of changes in this PR. High level overview:
- Extended nested walkers design, added extension points before and after the visit, some extra properties regarding the visit and fixed issues in visits of local methods 
- Made PXContext a field in nested walker and removed redundant pxcontext fields in inheritors
- Fixed bugs in PX1008 check
- Added check for captured adapters to PX1008 check and reworked diagnostic message
- Reworked PX1008 check completely. Now it is a data flow diagnostic that goes recursively into methods. It is based on new nested walkers design.
- Added support for PX1008 in custom views and attributes
- Added helper to check if  parameter is reassigned inside a method before some specified syntax node 
- Added helper that maps call arguments passed to the method to its parameters
- Added Nullable nuget package with attributes for nullable reference code analysis and added annotations to common utils
- Added many new helpers related to methods and extended existing ones
- Added tests for complex cases for PX1008

Here are some principles of the data flow analysis used by the main PX1008 walker.
We maintain a stack of non-capturable elements. Before we visit a methods from a method call recursively we calculate what  can capture reference to graph/adapter at the call site. The diagnostic supports four options:
- We are at the top of call stack and inside an action handler. In this case adapter parameter can be captured. 
- We are inside graph/graph extension and `this` is passed among the arguments
- We are inside recursive call and already have some non-capturable parameters that are used in call arguments. 
- We are inside a local method and parameters from containing methods have some non-capturable elements

 There is a dedicated walker that checks for parameters and `this` captured in arguments expression. 
Additionally, there is a walker that check for reassignments of parameters before the call site.

When we calculate these non-capturable elements we push them to the walker's stack. Then, when nested walker checks if it should go into the method, we look at the stack and go into a method only if something non-capturable is passed.
After the visit we remove the top of the stack.

One case is still left not covered by the data flow analysis - the passing of delegates that capture reference to graph like this:
```
CaptureGraph_PassedDelegate(graph.InstanceMember);
//...
private void CaptureGraph_PassedDelegate(PXToggleAsyncDelegate action)
{
	PXLongOperation.StartOperation(this, action);
}
```
It should be very rare and it is difficult to cover it properly since it would require to analyze lambda expressions


src/Acuminator/Acuminator.Analyzers/Resources.resx
```diff
@@ -166,7 +166,7 @@
     <value>The {0} DAC, DAC extension, or DAC property should have a description in the summary XML tag</value>
   </data>
   <data name="PX1008Title" xml:space="preserve">
-    <value>Reference of @this graph in this delegate will cause synchronous delegate execution. It may lead to random bugs and data consistency issues</value>
+    <value>The reference to {0} is captured in the {1} delegate closure and will cause synchronous delegate execution. It may lead to random bugs and data consistency issues</value>
```
SENya1990, line 169: @EAndrosova please review new resource strings for PX1008

dropsonic (approved):

dropsonic:
@SENya1990 kudos for writing such a beautiful and descriptive pull request explanation!

SENya1990:
@EAndrosova we will need to update the documentation for the PX1008 diagnostics because now we check for capturing in long run and processing delegates not only a reference to the screen graph but also a reference to  PXAdapter that is passed to action delegates

EAndrosova (approved):

PR #459, opened 2022-09-20
Bugfix/atr 347 dev px1050 string interpolation

SENya1990:
Review and merge only after https://github.com/Acumatica/Acuminator/pull/457

Changes overview:
- Added check of interpolation strings to PX1050 as part of the check for hard-coded strings
- Extended PX1050 diagnostic - added a new diagnostic message for the scenario when a string passed to the localization methods/exception constructor is not a constant string field.
- Removed flawed helper for argument to parameter mapping. Replaced it with arguments to parameters mapping utility. 
- Added unit tests for interpolation strings 
- Added symbol set for string type and refactored localization symbols
- Added some utils
- Refactored analyzers


src/Acuminator/Acuminator.Analyzers/Resources.resx
```diff
@@ -285,9 +285,12 @@
   <data name="PX1042Title" xml:space="preserve">
     <value>In a RowSelecting handler, BQL statements and other database queries must be executed only inside a separate connection scope</value>
   </data>
-  <data name="PX1050Title" xml:space="preserve">
+  <data name="PX1050Title_HardcodedString" xml:space="preserve">
     <value>Hardcoded strings are not allowed as parameters for localization methods and PXException constructors. You should use string constants from the appropriate localizable messages class.</value>
   </data>
+  <data name="PX1050Title_NonConstFieldString" xml:space="preserve">
+    <value>Strings that are not constant fields are not allowed as parameters for localization methods and PXException constructors. You should use string constants from the appropriate localizable messages class.</value>
```
SENya1990, line 292: @EAndrosova please review the new diagnostic message for the scenarios when string that is not a constant field is passed to the localization method or `PXException` constructor

dropsonic (approved):

PR #488, opened 2023-01-27
Bugfix/atr 718 dev px1095 false alert

SENya1990:
Finally finished DB Boundness calculation, it got much harder than I estimated initially due to tricky corner cases.

High level overview:
- Reworked completely work with Acumatica attributes - inheritance, flattening of aggregate attributes, data type and DB Boundness calculation. Redesigned and divided the old code which was a dump of different helpers without clear responsibilities and dependencies. Removed code duplication.
- Extended `DbBoundnessType` enum with new options - PXDBScalar , PXDBCalced , Error. Implemented rules for boundness combination for aggregate attributes and DAC properties. New values are integrated into calculations and Code Map.
- Separated calculation of DB boundness for attribute application on DAC property from the basic data type metadata.
- Added new kind of attribute metadata for attributes with mixed DB boundness that have flags `IsDBField` or `NonDB` (names hard-coded).
- Implemented flattening of all attributes' applications with aggregators support
- Implemented calculation of DB boundness with consideration of all explicitly set boundnesses like `IsDBField = true` or `NonDB = false` on all attribute applications including aggregates. The algorithm takes attribute application data:
             - Checks well-known non-data-type attributes like `PXUIFieldAttribute` for optimization
             - Gets flattened set of attributes applications including aggregated attributes and base types
             - Calculates data type attributes metadata from the flattened set
             - Combines explicit applications with metadata
             - If there is no defined DB boundness use duck typing as a last attempt. Look for `IsDBField` property in the flattened set of attributes. If there is none then the boundness is not defined, otherwise it is unknown.
- Added info about attribute metadata and flattened attributes set to the DAC semantic model
- Added unbound `PX.Data.PXVariantAttribute` to the data type attributes register with a data type `byte[]`. 
@dropsonic is `byte[]` the correct type? The [documentation ](https://help-2022r2.acumatica.com/(W(62))/Help?ScreenId=ShowWiki&pageid=92f5623d-eebb-e5e7-db17-5aa98633b219) and attribute usages are unclear, the attribute is used for both `byte[]` and `object` property types.
- Reworked analyzers and code fixes related to DAC properties analysis. In particular, the check for compatibility between DAC property type and attribute data type was reworked to use data from the DAC semantic model. Affected diagnostics:
          - https://github.com/Acumatica/Acuminator/blob/dev/docs/diagnostics/PX1023.md
          - https://github.com/Acumatica/Acuminator/blob/dev/docs/diagnostics/PX1021.md
          - https://github.com/Acumatica/Acuminator/blob/dev/docs/diagnostics/PX1019.md
          - https://github.com/Acumatica/Acuminator/blob/dev/docs/diagnostics/PX1020.md
          - https://github.com/Acumatica/Acuminator/blob/dev/docs/diagnostics/PX1030.md
          - https://github.com/Acumatica/Acuminator/blob/dev/docs/diagnostics/PX1095.md  
- Fixed and updated unit tests, refactored old tests for attribute helpers
- Added common helpers
- Mass renaming, replaced non-descriptive `special attributes` term with `attributes calced on DB side`
- Refactoring and nullable annotation
- Something else probably


dropsonic:
I'm not sure about `PXVariantAttribute`; I think the author is Kesha Burak.

dropsonic (approved):

SENya1990:
@dropsonic  according to Kesha it was Ilya Mochalov's creation. DAC property can be weekly typed with `object`, however, in general it's expected to be `byte[]` to work correctly. It's unknown if other cases are supported, so it's safer to assume that DAC property with this attribute should have type `byte[]`

dropsonic:
> @dropsonic according to Kesha it was Ilya Mochalov's creation. DAC property can be weekly typed with `object`, however, in general it's expected to be `byte[]` to work correctly. It's unknown if other cases are supported, so it's safer to assume that DAC property with this attribute should have type `byte[]`

Good.
Personally, I think that using `PXVariantAttribute` is a bad practice. It is almost impossible to use such fields in DB migrations.

PR #495, opened 2023-05-09
PXOverride Analyzer - initial commit.

taskinkemal:
PXOverride Signature mismatch analyzer.
Provides two diagnostics:
1) Signature check
2) PXOverride usage in a class that is not inherited from a PXGraphExtension.


SENya1990:
@taskinkemal I have a remark about the commits policy. Please use more granular git commits in the future. It really helps to investigate the history of some code snippet when you have a detailed history of its commits. And it is hard to do an investigation if all changes are merged together into a single commit. Please try to commit together only related changes with the commit message describing what was changed.

src/Acuminator/Acuminator.Analyzers/DiagnosticsShortName.resx
```diff
@@ -384,4 +384,10 @@
   <data name="PX1095PXDBScalar" xml:space="preserve">
     <value>NoUnboundTypeAttributeWithPXDBScalar</value>
   </data>
+  <data name="PX1096ContainerTypeError" xml:space="preserve">
```
SENya1990, line 387: @EAndrosova  please review the new diagnostic short names

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/Descriptors.cs
```diff
@@ -412,5 +412,11 @@ public static class Descriptors
 
 		public static DiagnosticDescriptor PX1095_PXDBScalarMustBeAccompaniedNonDBTypeAttribute { get; } =
 			Rule("PX1095", nameof(Resources.PX1095Title_PXDBScalar).GetLocalized(), Category.Default, DiagnosticSeverity.Error, DiagnosticsShortName.PX1095PXDBScalar);
+
```
SENya1990, line 415: The severity of both diagnostic should be Error

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMismatchAnalyzer.cs
```diff
@@ -0,0 +1,230 @@
...
+using Microsoft.CodeAnalysis.Diagnostics;
+using System.Linq;
+using Acuminator.Analyzers.StaticAnalysis.AnalyzersAggregator;
+using Acuminator.Utilities;
+using Acuminator.Utilities.DiagnosticSuppression;
+
+namespace Acuminator.Analyzers.StaticAnalysis.PXOverrideMismatch
+{
+	[DiagnosticAnalyzer(LanguageNames.CSharp)]
+	public class PXOverrideMismatchAnalyzer : SymbolAnalyzersAggregator<IPXGraphAnalyzer>
```
SENya1990, line 15: @taskinkemal this is incorrect design for the analyzer. The `SymbolAnalyzersAggregator` should be used only of you are looking to aggregate several analyzers that can share a common state between them such as graph and dac analyzers we have. That's why we have a notion of the "aggregated" analyzers that are not actually derived from the Roslyn base analyzer class. 

In case of this diagnostic, if you want to check the `PXOverride` signature you need to analyze only graph extensions and we have a perfect base class for that. You should derive from the `PXGraphAggregatedAnalyzerBase` and implement abstract base classes. 

Take a look at `CallingBaseActionHandlerFromOverrideHandlerAnalyzer` and `NoIsActiveMethodForExtensionAnalyzer` analyzers for an example

SENya1990 (changes requested):
There are significant flaws in the analyzer design that should be addressed first

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/Descriptors.cs
```diff
@@ -412,5 +412,11 @@ public static class Descriptors
...
+		public static DiagnosticDescriptor PX1096_PXOverrideMustMatchSignature { get; } =
+			Rule("PX1096", nameof(Resources.PX1096Title_SignatureMismatch).GetLocalized(), Category.Default, DiagnosticSeverity.Warning, DiagnosticsShortName.PX1096SignatureMismatch);
+
+		public static DiagnosticDescriptor PX1096_PXOverrideContainerIsNotPXGraphExtension { get; } =
```
SENya1990, line 419: I'm not sure that this check is needed. We don't add checks for everything just because it can be checked since every diagnostic takes developer machine resources and slows down IDE a bit. 

I will discuss it with the ISV team but I would advise to remove this check. Supporting PXOverride lookup in every class will force you to design an analyzer that have to check type declarations in code instead of graph extensions only. This is bad for performance. If this scenario is really useful and encountered by our developers then it's still better to split the analyzer into two analyzers - an aggregated graph analyzer that will check only graph extensions for the correct signature and the standard Acumiantor roslyn analyzer derived from `PXDiagnosticAnalyzer` that will check the compilation root node, collect all "PXOverride" attributes from it and check its containing types.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMismatchAnalyzer.cs
```diff
@@ -0,0 +1,230 @@
...
+			var graphType = FindFirstTypeArgumentOfPxGraphExtension(extensionType);
+
+			if (graphType != null)
+			{
+				GetBaseTypes(pxGraphExtensionTypes, graphType, hashSet);
+			}
+
+			GetBaseTypes(pxGraphExtensionTypes, typeSymbol.BaseType, hashSet);
+		}
+		private static INamedTypeSymbol FindFirstTypeArgumentOfPxGraphExtension(INamedTypeSymbol type)
```
SENya1990, line 113: We have a number of extension helpers that already implement these utils in the
`Acuminator.Utilities.Roslyn.Semantic.PXGraph`. This code should be reworked by using those utils and these helpers should be removed

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMismatchAnalyzer.cs
```diff
@@ -0,0 +1,230 @@
...
+				// PXOverride can only be used in a PXGraphExtension child type.
+				var diagnostic = Diagnostic.Create(
+					Descriptors.PX1096_PXOverrideContainerIsNotPXGraphExtension,
+					methodSymbol.Locations.First());
+
+				context.ReportDiagnosticWithSuppressionCheck(diagnostic, pxContext.CodeAnalysisSettings);
+			}
+		}
+
+		private static HashSet<INamedTypeSymbol> GetAllBaseTypes(
```
SENya1990, line 78: We already have utility for all these helpers. They must be removed

src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/PXContext.cs
```diff
@@ -93,6 +94,10 @@ public class PXContext
 		public PXGraphExtensionSymbols PXGraphExtension => _pxGraphExtensionSymbols.Value;
 
 		public INamedTypeSymbol PXCacheExtensionType => Compilation.GetTypeByMetadataName(TypeFullNames.PXCacheExtension);
+		public INamedTypeSymbol PXGraphExtensionType => Compilation.GetTypeByMetadataName(TypeFullNames.PXGraphExtension);
```
SENya1990, line 97: We already have PXGraphExtension type just few lines above. Please remove this duplicate. 

As for the generic types - they should be removed too. We can't use just a couple of generic types, we must support all generic graph extension types, there could be high level chained graph extensions that have much more type parameters. It's not an option to hard code the types, so they should be also removed.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMismatchAnalyzer.cs
```diff
@@ -0,0 +1,230 @@
...
+			List<INamedTypeSymbol> pxGraphExtensionTypes,
+			INamedTypeSymbol typeSymbol,
+			HashSet<INamedTypeSymbol> hashSet)
+		{
+			if (typeSymbol == null || typeSymbol.SpecialType != SpecialType.None || hashSet.Contains(typeSymbol))
+			{
+				return;
+			}
+
+			hashSet.Add(typeSymbol);
```
SENya1990, line 101: hashset Add and hashset Contains should be combined into a single call

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMismatchAnalyzer.cs
```diff
@@ -0,0 +1,230 @@
...
+
+			foreach (var type in types)
+			{
+				GetBaseTypes(pxGraphExtensionTypes, type, allTypes);
+			}
+
+			return allTypes;
+		}
+		private static void GetBaseTypes(
+			List<INamedTypeSymbol> pxGraphExtensionTypes,
```
SENya1990, line 92: please use meaningful names for every code element. Names like `hashset` do not tell anything to the reader about what it really is, and how it is used. I can see that it is a hashset just from the parameter type, so there is no need to duplicate this info.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMismatchAnalyzer.cs
```diff
@@ -0,0 +1,230 @@
...
+			if (baseTypes.Any(b => typeSymbol.MetadataName == b.MetadataName))
+			{
+				return true;
+			}
+
+			return Implements(typeSymbol.BaseType, baseTypes);
+		}
+
+
+		private static bool IsCompatibleMethod(IMethodSymbol src, IMethodSymbol ext)
```
SENya1990, line 159: IsCompatible method doesn't check if the method from the base type has declared accessibility `public`/`protected`, nor does it check if the method is `virtual`.  It also doesn't check method names.

Please use sensible names for parameters. And it could be good to refactor this method by splitting checks into several methods. I think you can reuse the parameters comparison which is done twice by making it a separate method.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMismatchAnalyzer.cs
```diff
@@ -0,0 +1,230 @@
...
+			{
+				return type.TypeArguments.FirstOrDefault() as INamedTypeSymbol;
+			}
+
+			return FindFirstTypeArgumentOfPxGraphExtension(type.ContainingType);
+		}
+
+		private static INamedTypeSymbol GetPXGraphExtension(List<INamedTypeSymbol> pxGraphExtensionTypes, INamedTypeSymbol typeSymbol)
+		{
+			if (typeSymbol == null)
```
SENya1990, line 130: No need for a null check at this point

SENya1990 (changes requested):
Added few more remarks, the analyzer still should be reworked

src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/SymbolEqualityComparer.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System.Collections.Generic;
+using Microsoft.CodeAnalysis;
+
+namespace Acuminator.Utilities.Roslyn.Semantic
+{
+	public sealed class SymbolEqualityComparer : IEqualityComparer<ISymbol>
```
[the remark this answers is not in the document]
taskinkemal: This is true, I was porting back from a newer version so wanted to leave a trace for the future, so it'd be easier when you want to do this update.

I'll do this change. Thank you.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMismatchAnalyzer.cs
taskinkemal: Thank you for your feedback. This method can be made cleaner, this is true. We can also add unit tests around it. I'll work on it.

One thing on the method name: I think Acumatica allows different method names to be px-overriden. We have examples on our code base.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/Descriptors.cs
SENya1990: @taskinkemal I've discussed with ISVs the `PXOverride` scenario, it seems that this check can be useful for them. In this case it should be done in a separate analyzer.

src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/PXContext.cs
taskinkemal: Removed.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMismatchAnalyzer.cs
taskinkemal: This seems to be a specific implementation. I'm not sure it'd be preferrable to move this to the utility class. Please let me know.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/Descriptors.cs
taskinkemal: With respect to another comment above, maybe it's best to exclude this rule from this commit and keep it as simple as possible. Later we can add this analyzer as a separate commit.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMismatchAnalyzer.cs
taskinkemal: Ok, this is done. Thank you for your feedback. Please review again.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/Descriptors.cs
taskinkemal: Done, thank you.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMismatchAnalyzer.cs
taskinkemal: This might be a mistake from my side. I rerun the analyzer on our code base with the name constraint and it seems to be fine. I'll add the name check. Thank you.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMismatchAnalyzer.cs
taskinkemal: This is a recursive method and it might still make sense as a base condition. Please let me know.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMismatchAnalyzer.cs
taskinkemal: This is also done, thank you.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMismatchAnalyzer.cs
taskinkemal: This variable is renamed now.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/Descriptors.cs
SENya1990: I'm not sure I understand what you mean by commit. If you mean a separate PR with implementation of that analyzer than yes, I agree. It would be cleaner to have a separate PR for that although you would probably need to remove related unit tests from this PR.

If you are talking about just a separate git commits inside this branch then please make more granular git commits. It shouldn't be one git commit per PR.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMismatchAnalyzer.cs
SENya1990: Actually, I think that this method is also covered by existing helpers

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideExtensions.cs
```diff
@@ -0,0 +1,81 @@
+﻿using System.Collections.Generic;
```
SENya1990, line 1: For new code we use C# 8 nullable ref types and annotate code accordingly:
```c#
#nullable enable
```

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideExtensions.cs
```diff
@@ -0,0 +1,81 @@
...
+
+			if (type.TypeArguments.Any())
+			{
+				return type.TypeArguments.FirstOrDefault() as INamedTypeSymbol;
+			}
+
+			return type.ContainingType.GetFirstTypeArgument();
+		}
+
+		public static INamedTypeSymbol GetPXGraphExtension(this INamedTypeSymbol typeSymbol, INamedTypeSymbol pxGraphExtensionType)
```
SENya1990, line 26: I have several remarks about this helper.
1. We keep public utility methods that could be reused in the Utility project in a suitable namespace and class. 
2. We throw argument null exception if the input is null. No need to support case with null argument.
3. This check spends too much resources. There is no need for recursive check here which will call inherits helper which already checks the whole type hierarchy. 
4. If what you are trying to obtain is PXGraphExtension<TBaseExxtension, ...., , TGraph> then this code is not correct too. It doesn't cover cases like this:
```csharp
public class SomeExtension<TGraph, TDac> : PXGraphExtension<TGraph> {}
```
It is generic and is inherited from `PXGraphExtension`

Overall, I don't see a need in this method if you will redesign analyzer to be the aggregated graph analyzer

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMethodSymbolComparer.cs
```diff
@@ -0,0 +1,107 @@
...
+using System.Linq;
+using Acuminator.Utilities.Common;
+
+namespace Acuminator.Analyzers.StaticAnalysis.PXOverrideMismatch
+{
+	public class PXOverrideMethodSymbolComparer : IEqualityComparer<IMethodSymbol>
+	{
+		public static PXOverrideMethodSymbolComparer Instance = new PXOverrideMethodSymbolComparer();
+
+		public bool Equals(IMethodSymbol src, IMethodSymbol ext)
```
SENya1990, line 13: please use full names, no shortcuts. I would propose `sourceMethod` and `pxOverrideMethod`.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMethodSymbolComparer.cs
```diff
@@ -0,0 +1,107 @@
...
+		{
+			var srcParams = src.Parameters.Select(p => p.Type).ToImmutableArray();
```
SENya1990, line 15: We have extra requirements for the performance since this code is running constantly in the IDE on the developer machine. This method can be made more performant:
You always do LINQ transformation and then always buffer it into the Immutable array even if there is no need to do it.  You can check parameters length before any need to use any parameter type. Moreover, this conversion is not needed at all. No need to allocate two more collections when you can work directly with parameters collections.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMethodSymbolComparer.cs
```diff
@@ -0,0 +1,107 @@
...
+			var extParams = ext.Parameters.Select(p => p.Type).ToImmutableArray();
+
+			var isSimple = srcParams.Length == extParams.Length;
```
SENya1990, line 18: Please use more sensible names. It's not good that Acumatica PXOverride resolution logic have not sensible names but we have a stricter code quality policy for Acuminator.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMethodSymbolComparer.cs
```diff
@@ -0,0 +1,107 @@
...
+using System.Collections.Immutable;
+using Microsoft.CodeAnalysis;
+using System.Linq;
+using Acuminator.Utilities.Common;
+
+namespace Acuminator.Analyzers.StaticAnalysis.PXOverrideMismatch
+{
+	public class PXOverrideMethodSymbolComparer : IEqualityComparer<IMethodSymbol>
+	{
+		public static PXOverrideMethodSymbolComparer Instance = new PXOverrideMethodSymbolComparer();
```
SENya1990, line 11: This should be readonly

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMethodSymbolComparer.cs
```diff
@@ -0,0 +1,107 @@
...
+
+			var isSimple = srcParams.Length == extParams.Length;
+			var isComplex = srcParams.Length + 1 == extParams.Length;
+
+			if (!isSimple && !isComplex)
+			{
+				return false;
+			}
+
+			if (!HasCorrectOverrideSignature(src) || !HasCorrectAccessibility(src))
```
SENya1990, line 26: The original method could be `protected` but the  pxoverride method must be public.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMethodSymbolComparer.cs
```diff
@@ -0,0 +1,107 @@
+﻿using System.Collections.Generic;
+using System.Collections.Immutable;
+using Microsoft.CodeAnalysis;
+using System.Linq;
+using Acuminator.Utilities.Common;
+
+namespace Acuminator.Analyzers.StaticAnalysis.PXOverrideMismatch
+{
+	public class PXOverrideMethodSymbolComparer : IEqualityComparer<IMethodSymbol>
```
SENya1990, line 9: I don't recommend to move out the comparison logic out of the analyzer. It's not wrong to have a separate component for it but in the future we will most likely enhance the checks and report different cases of partial compatibility between the method, so there will be extra analysis logic added in the future which won't fit well into `IEqualityComparer` contract.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMethodSymbolComparer.cs
SENya1990: You can override in a graph extension not generic public virtual or not generic protected virtual methods of the graph/lower level chained extensions. Thus,  you need to check that the original method is virtual and that both methods do not introduce their own generic type parameters.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMethodSymbolComparer.cs
```diff
@@ -0,0 +1,107 @@
...
+		private static bool HasCorrectOverrideSignature(IMethodSymbol methodSymbol) => methodSymbol.IsVirtual || methodSymbol.IsOverride;
+
+		private static bool HasCorrectAccessibility(IMethodSymbol methodSymbol) =>
+			methodSymbol.DeclaredAccessibility == Accessibility.Public ||
+			methodSymbol.DeclaredAccessibility == Accessibility.Protected ||
+			methodSymbol.DeclaredAccessibility == Accessibility.ProtectedOrInternal;
+
+		private static bool IsDelegateCompatible(ITypeSymbol srcReturnType, ImmutableArray<ITypeSymbol> srcParameters, INamedTypeSymbol targetDelegate)
+		{
+			if (!targetDelegate.IsGenericType && targetDelegate.TypeKind != TypeKind.Delegate)
```
SENya1990, line 65: what does this check mean? Why non generic delegates are disallowed? It's not correct.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMethodSymbolComparer.cs
```diff
@@ -0,0 +1,107 @@
...
+			{
+				return false;
+			}
+
+			var returnType = targetDelegate.TypeKind == TypeKind.Delegate ? targetDelegate.DelegateInvokeMethod.ReturnType : targetDelegate.TypeArguments.Last();
```
SENya1990, line 70: What generic non delegate type is supported here?

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/Descriptors.cs
taskinkemal: That's true. Sorry, I meant PR. I've removed the related unit tests already. Thank you.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideExtensions.cs
taskinkemal: Ok, sure, I'll use nullable types then.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMismatchAnalyzer.cs
```diff
@@ -0,0 +1,63 @@
+﻿using System.Collections.Immutable;
```
SENya1990, line 1: Please add `nullable enable`

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMismatchAnalyzer.cs
```diff
@@ -0,0 +1,63 @@
...
+				.Where(m => !m.IsStatic && m.HasAttribute(pxContext.AttributeTypes.PXOverrideAttribute, false))
+				.ForEach(m => AnalyzeMethod(context, pxContext, m));
+		}
+
+		private void AnalyzeMethod(SymbolAnalysisContext context, PXContext pxContext, IMethodSymbol methodSymbol)
+		{
+			// Here we know that the method is not static and has the correct attribute.
+
+			var extensionType = methodSymbol.ContainingType.GetPXGraphExtension(pxContext.PXGraphExtension.Type);
+			var graphType = extensionType.GetFirstTypeArgument();
```
SENya1990, line 38: The helper is incorrect like I wrote before. Moreover it is not needed, Pass the graph semantic model to this method, it has a separate graph symbol

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMismatchAnalyzer.cs
```diff
@@ -0,0 +1,63 @@
...
+
+			if (graphType != null)
+			{
+				context.CancellationToken.ThrowIfCancellationRequested();
+
+				var allTypes = PXOverrideExtensions.GetAllBaseTypes(pxContext.PXGraphExtension.Type, extensionType, graphType);
+
+				context.CancellationToken.ThrowIfCancellationRequested();
+
+				var hasMatchingMethod = allTypes
```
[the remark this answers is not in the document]
taskinkemal: Thank you for this. I'll change the code with a similar approach.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMismatchAnalyzer.cs
```diff
@@ -0,0 +1,63 @@
...
+			// Here we know that the method is not static and has the correct attribute.
+
+			var extensionType = methodSymbol.ContainingType.GetPXGraphExtension(pxContext.PXGraphExtension.Type);
+			var graphType = extensionType.GetFirstTypeArgument();
+
+			if (graphType != null)
+			{
+				context.CancellationToken.ThrowIfCancellationRequested();
+
+				var allTypes = PXOverrideExtensions.GetAllBaseTypes(pxContext.PXGraphExtension.Type, extensionType, graphType);
```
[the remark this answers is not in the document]
taskinkemal: Actually there are two cases I need to confirm with you. If those are ok, then I can use this one. And that helper class can be removed as well. When I use this method TypeArgumentImplementsPxGraphExtension and PxOverrideInADifferentType tests fail. Let me know if this is ok. We can check tomorrow.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMismatchAnalyzer.cs
taskinkemal: Done, thanks.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMismatchAnalyzer.cs
```diff
@@ -0,0 +1,63 @@
...
+
+				var hasMatchingMethod = allTypes
+					.SelectMany(t => t.GetMethods(methodSymbol.Name))
+					.Any(m => PXOverrideMethodSymbolComparer.Instance.Equals(m, methodSymbol));
+
+				if (!hasMatchingMethod)
+				{
+					var diagnostic = Diagnostic.Create(
+						Descriptors.PX1096_PXOverrideMustMatchSignature,
+						methodSymbol.Locations.First());
```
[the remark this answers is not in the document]
SENya1990: Please notify me on update for the remarks that have been fixed, so I can check them and resolve the conversation

src/Acuminator/Acuminator.Analyzers/DiagnosticsShortName.resx
```diff
@@ -384,4 +384,10 @@
   <data name="PX1095PXDBScalar" xml:space="preserve">
     <value>NoUnboundTypeAttributeWithPXDBScalar</value>
   </data>
+  <data name="PX1097" xml:space="preserve">
```
SENya1990, line 387: Please remove

src/Acuminator/Acuminator.Analyzers/Resources.resx
```diff
@@ -670,4 +670,10 @@
   <data name="PX1095Title_PXDBScalar" xml:space="preserve">
     <value>A field with the PXDBScalar attribute must have an unbound type attribute, such as PXDate, PXDecimal</value>
   </data>
+  <data name="PX1097Title" xml:space="preserve">
```
SENya1990, line 673: please remove

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/Descriptors.cs
```diff
@@ -412,5 +412,11 @@ public static class Descriptors
 
 		public static DiagnosticDescriptor PX1095_PXDBScalarMustBeAccompaniedNonDBTypeAttribute { get; } =
 			Rule("PX1095", nameof(Resources.PX1095Title_PXDBScalar).GetLocalized(), Category.Default, DiagnosticSeverity.Error, DiagnosticsShortName.PX1095PXDBScalar);
+
+		public static DiagnosticDescriptor PX1096_PXOverrideMustMatchSignature { get; } =
```
SENya1990, line 416: please remove

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMethodSymbolComparer.cs
```diff
@@ -0,0 +1,93 @@
...
+			{
+				return false;
+			}
+
+			return DelegateSignatureMatches(sourceMethod.ReturnType, sourceMethodParameterTypes, @delegate);
+		}
+
+		private static bool HasCorrectOverrideSignature(IMethodSymbol methodSymbol) => methodSymbol.IsVirtual || methodSymbol.IsOverride;
+
+		private static bool HasCorrectAccessibility(IMethodSymbol methodSymbol) =>
```
SENya1990, line 68: IsAccessibleOutsideOfAssembly

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMethodSymbolComparer.cs
```diff
@@ -0,0 +1,93 @@
...
+		{
+			var isSimple = sourceMethod.Parameters.Length == extensionMethod.Parameters.Length;
+			var isComplex = sourceMethod.Parameters.Length + 1 == extensionMethod.Parameters.Length;
+
+			if (!isSimple && !isComplex)
+			{
+				return false;
+			}
+
+			if (!HasCorrectOverrideSignature(sourceMethod) || !HasCorrectAccessibility(sourceMethod))
```
SENya1990, line 35: Check if the method is generic

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMethodSymbolComparer.cs
```diff
@@ -0,0 +1,93 @@
...
+
+			var sourceMethodParameterTypes = sourceMethod.Parameters.Select(p => p.Type);
+			var extensionMethodParameterTypes = extensionMethod.Parameters.Select(p => p.Type);
+
+			if (isSimple)
+			{
+				return sourceMethodParameterTypes.SequenceEqual(extensionMethodParameterTypes);
+			}
+
+			if (extensionMethodParameterTypes.Last() is not INamedTypeSymbol @delegate)
```
SENya1990, line 53:
```suggestion
			if (extensionMethodParameterTypes.Last() is not INamedTypeSymbol @delegate || @delegate.TypeKind !=TypeKind.Delegate )
```

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMethodSymbolComparer.cs
```diff
@@ -0,0 +1,93 @@
...
+
+			if (!sourceMethodParameterTypes.SequenceEqual(extensionMethodParameterTypes.Take(extensionMethod.Parameters.Length - 1)))
+			{
+				return false;
+			}
+
+			return DelegateSignatureMatches(sourceMethod.ReturnType, sourceMethodParameterTypes, @delegate);
+		}
+
+		private static bool HasCorrectOverrideSignature(IMethodSymbol methodSymbol) => methodSymbol.IsVirtual || methodSymbol.IsOverride;
```
SENya1990, line 66: please move to IMethodSymbolExtensions

src/Acuminator/Acuminator.Tests/Tests/StaticAnalysis/PXOverrideMismatch/Sources/PxOverrideInADifferentType.cs
```diff
@@ -0,0 +1,56 @@
...
+	public class ExtClass : BaseClass
+	{
+		[PX.Data.PXOverride]
+		public override object TestMethod(int x, bool drilldown, double y)
+		{
+			return new object();
+		}
+	}
+
+	public class MyGraph : PX.Data.PXGraph
```
SENya1990, line 22:
```suggestion
	public class MyGraph : PX.Data.PXGraph<MyGraph>
```

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMismatchAnalyzer.cs
taskinkemal: This is removed as requested. Thank you.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMismatchAnalyzer.cs
taskinkemal: We use GetGraphExtensionWithBaseExtensions now, the helpers are removed. Thank you.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideExtensions.cs
taskinkemal: We use GetGraphExtensionWithBaseExtensions now, these helpers are removed. Thank you!

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMethodSymbolComparer.cs
taskinkemal: Names are reworked now. Thank you for your help.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMethodSymbolComparer.cs
taskinkemal: Both checks are done now.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMethodSymbolComparer.cs
```diff
@@ -0,0 +1,107 @@
...
+			{
+				return false;
+			}
+
+			if (!HasCorrectOverrideSignature(src) || !HasCorrectAccessibility(src))
+			{
+				return false;
+			}
+
+			if (!src.ReturnType.Equals(ext.ReturnType))
```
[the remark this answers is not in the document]
taskinkemal: Thank you for your feedback. This check is moved up. And in general the more expensive checks are pushed down.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMismatchAnalyzer.cs
```diff
@@ -0,0 +1,63 @@
...
+
+		public override bool ShouldAnalyze(PXContext pxContext, PXGraphSemanticModel graph) =>
+			base.ShouldAnalyze(pxContext, graph) && graph.Type == GraphType.PXGraphExtension;
+
+		public override void Analyze(SymbolAnalysisContext context, PXContext pxContext, PXGraphSemanticModel pxGraphExtension)
+		{
+			context.CancellationToken.ThrowIfCancellationRequested();
+
+			pxGraphExtension.Symbol
+				.GetMembers()
```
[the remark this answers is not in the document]
taskinkemal: This is done. Thank you.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMismatchAnalyzer.cs
taskinkemal: We use GetGraphExtensionWithBaseExtensions now, these helpers are removed. Thank you!

src/Acuminator/Acuminator.Analyzers/DiagnosticsShortName.resx
taskinkemal: Removed.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMethodSymbolComparer.cs
```diff
@@ -0,0 +1,93 @@
...
+		/// In both cases, PXOverride dictates that the base method must be
+		///     1) virtual (having either a virtual or an override signature),
+		///     2) public, protected or protected internal.
+		/// </summary>
+		/// <param name="sourceMethod">The method from the base</param>
+		/// <param name="extensionMethod">The method from the derived class, with the PXOverride attribute</param>
+		/// <returns></returns>
+		public static bool Equals(IMethodSymbol sourceMethod, IMethodSymbol extensionMethod)
+		{
+			var isSimple = sourceMethod.Parameters.Length == extensionMethod.Parameters.Length;
```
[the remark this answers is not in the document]
taskinkemal: This is done. Thank you.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMethodSymbolComparer.cs
taskinkemal: This is done. IsGeneric check is done very early now.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMethodSymbolComparer.cs
taskinkemal: This is done. Thank you.

src/Acuminator/Acuminator.Tests/Tests/StaticAnalysis/PXOverrideMismatch/Sources/PxOverrideInADifferentType.cs
taskinkemal: The last commit contains a code cleanup in the tests: naming, removal of the redundant PX.Data namespace, and PXGraph<TGraph>. Please let me know if it looks good.

taskinkemal:
I think the feedback from the code reviews are all covered. I've also updated the documentation. I'll continue working on this if there is further feedback. Thank you for your time and attention.

docs/Summary.md
```diff
@@ -93,6 +93,7 @@ Acuminator does not perform static analysis of projects whose names contain `Tes
 | [PX1093](diagnostics/PX1093.md) | In a graph declaration, the first type parameter of `PXGraph` must be the graph type. | Error | Available |
 | [PX1094](diagnostics/PX1094.md) | Every DAC should be decorated with the `PXHidden` or `PXCacheName` attribute. | Warning (ISV Level 3: Informational) | Available |
 | [PX1095](diagnostics/PX1095.md) | A field with the `PXDBCalced` or `PXDBScalar` attribute must have an unbound type attribute, such as `PXDate` or `PXDecimal`. | Error | Unavailable |
+| [PX1096](diagnostics/PX1096.md) | Signature of a method with the `PXOverride` attribute must match the overriden method. | Error | Unavailable |
```
SENya1990, line 96: @EAndrosova please check the provided summary for the new diagnostic PX1096

docs/diagnostics/PX1096.md
```diff
@@ -0,0 +1,107 @@
+# PX1096
+This document describes the PX1096 diagnostic.
```
SENya1990, line 2: @EAndrosova please check the provided documentation for the new diagnostic PX1096

src/Acuminator/Acuminator.Analyzers/DiagnosticsShortName.resx
```diff
@@ -384,4 +384,7 @@
   <data name="PX1095PXDBScalar" xml:space="preserve">
     <value>NoUnboundTypeAttributeWithPXDBScalar</value>
   </data>
+  <data name="PX1096" xml:space="preserve">
+    <value>PXOverrideSignatureMismatch</value>
```
SENya1990, line 388: @EAndrosova please check the short name for the new diagnostic

src/Acuminator/Acuminator.Analyzers/Resources.resx
```diff
@@ -670,4 +670,7 @@
   <data name="PX1095Title_PXDBScalar" xml:space="preserve">
     <value>A field with the PXDBScalar attribute must have an unbound type attribute, such as PXDate, PXDecimal</value>
   </data>
+  <data name="PX1096Title" xml:space="preserve">
+    <value>PXOverride method signature must match the overriden method</value>
```
SENya1990, line 674: @EAndrosova please check the error message for the new diagnostic

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMismatchAnalyzer.cs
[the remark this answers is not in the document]
SENya1990: @taskinkemal I have investigated the case when there is a derived graph extension that both overrides and PXOverrides method from the base extension. The case doesn't have any clear useful scenario and the  platform does not support such case. Which means that analysis of suitable base methods for PXOverride should not found such method.

Regarding the second case - when you add .Net inheritance, especially with non-abstract, non-generic extension classes in the middle, the graph extensions resolution becomes very complicated. If there is no such case in the Lexware codebase then I suggest to skip this case for now.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMismatchAnalyzer.cs
```diff
@@ -0,0 +1,166 @@
...
+		{
+			context.CancellationToken.ThrowIfCancellationRequested();
+
+			var allMembers = pxGraphExtension.Symbol.GetMembers();
+
+			if (!allMembers.IsDefaultOrEmpty)
+			{
+				var methodsWithPxOverrideAttribute = allMembers
+					.OfType<IMethodSymbol>()
+					.Where(m => !m.IsStatic && !m.IsGenericMethod && m.HasAttribute(pxContext.AttributeTypes.PXOverrideAttribute, false))
```
SENya1990, line 35: Please use named parameters with literal values such as `false` for readability

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMismatchAnalyzer.cs
```diff
@@ -0,0 +1,166 @@
...
+					.ToList();
+
+				if (methodsWithPxOverrideAttribute.Any())
+				{
+					var allBaseTypes = pxGraphExtension.Symbol
+						.GetGraphExtensionWithBaseExtensions(pxContext, SortDirection.Ascending, true)
```
SENya1990, line 41: Please use named parameters with literal values such as `true` for readability

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMismatchAnalyzer.cs
```diff
@@ -0,0 +1,166 @@
...
+			public static bool IsSuitable(IMethodSymbol baseMethod, IMethodSymbol pxOverrideMethod)
+			{
+				var methodsCompatibility = GetMethodsCompatibility(baseMethod.Parameters.Length, pxOverrideMethod.Parameters.Length);
+
+				if (methodsCompatibility == MethodsCompatibility.NotCompatible)
+				{
+					return false;
+				}
+
+				if (!pxOverrideMethod.CanBeOverriden() || !pxOverrideMethod.IsAccessibleOutsideOfAssembly())
```
SENya1990, line 96: I believe this should be the check for the `baseMethod`
```suggestion
				if (!baseMethod.CanBeOverriden() || !baseMethod.IsAccessibleOutsideOfAssembly())
```

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMismatchAnalyzer.cs
```diff
@@ -0,0 +1,166 @@
...
+				}
+
+				return true;
+			}
+
+			private static bool DoParametersMatch(ImmutableArray<IParameterSymbol> sourceParameters, ImmutableArray<IParameterSymbol> targetParameters)
+			{
+				for (var i = 0; i < sourceParameters.Length; i++)
+				{
+					if (i >= targetParameters.Length || !sourceParameters[i].Type.Equals(targetParameters[i].Type))
```
SENya1990, line 143: the check `i >= targetParameters.Length` is unnecessary at this stage. We should have already filtered all suitable base method candidates that have more parameters then PXOverride method. No need to check it, especially in a cycle:
```suggestion
					if (!sourceParameters[i].Type.Equals(targetParameters[i].Type))
```

src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/IMethodSymbolExtensions.cs
```diff
@@ -347,5 +347,13 @@ public static bool IsDefinitelyStatic(this IMethodSymbol method, SyntaxNode meth
 
 			return methodDeclaration.IsStatic();
 		}
+
+		/// <summary>
+		/// Check if <paramref name="method"/> has either virtual or override signature.
+		/// </summary>
+		/// <param name="method">The method to act on.</param>
+		/// <returns>True if <paramref name="method"/> </returns>
+		public static bool CanBeOverriden(this IMethodSymbol method) => method.IsVirtual || method.IsOverride;
```
SENya1990, line 356: It's better to check for null in public APIs, but I don't insist on this one

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMismatchAnalyzer.cs
taskinkemal: Done.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMismatchAnalyzer.cs
taskinkemal: Done.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMismatchAnalyzer.cs
taskinkemal: Thank you for your feedback. This is fixed.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMismatchAnalyzer.cs
taskinkemal: Removed. Thank you.

src/Acuminator/Acuminator.Analyzers/StaticAnalysis/PXOverrideMismatch/PXOverrideMismatchAnalyzer.cs
```diff
@@ -0,0 +1,167 @@
...
+			if (!allMembers.IsDefaultOrEmpty)
+			{
+				var methodsWithPxOverrideAttribute = allMembers
+					.OfType<IMethodSymbol>()
+					.Where(m => m.HasAttribute(pxContext.AttributeTypes.PXOverrideAttribute, checkOverrides: false))
+					.ToList();
+
+				if (methodsWithPxOverrideAttribute.Any())
+				{
+					var allBaseTypes = pxGraphExtension.Symbol
```
SENya1990, line 40: After discussion with colleagues about the PXOverride behavior in the system it seems that we should add a bit extra logic to exclude direct base types of the extension from the list of all base types in which we search for the PXOverride. 

Imho it should be something like this:
```csharp
if (methodsWithPxOverrideAttribute.Any())
{
      var directBaseTypesAndThis = pxGraphExtension.Symbol.GetBaseTypesAndThis().ToList(capacity: 4);
      var allBaseTypes = pxGraphExtension.Symbol
                                   .GetGraphExtensionWithBaseExtensions(pxContext, SortDirection.Ascending, includeGraph: true)
				  .OfType<INamedTypeSymbol>()
				  .Where(baseType => !directBaseTypesAndThis .Contains(baseType));

     //...
}
```

The rest of PR now looks good to me

taskinkemal: Changes are pushed.

SENya1990 (approved):
Everything looks good for me now

src/Acuminator/Acuminator.Utilities/Roslyn/Semantic/IMethodSymbolExtensions.cs
```diff
@@ -347,5 +347,18 @@ public static bool IsDefinitelyStatic(this IMethodSymbol method, SyntaxNode meth
...
+		/// <summary>
+		/// Check if <paramref name="method"/> has either virtual or override signature.
+		/// </summary>
+		/// <param name="method">The method to act on.</param>
+		/// <returns>True if <paramref name="method"/> </returns>
+		public static bool CanBeOverriden(this IMethodSymbol method)
+		{
+			method.ThrowOnNull(nameof(method));
+
+			return method.IsVirtual || method.IsOverride;
```
dropsonic, line 360: `abstract` members can also be overridden, should we also check for it?

SENya1990: Yes. I checked and Roslyn considers abstract methods to be non virtual, so we should check `IsAbstract` flag too. Thanks for noticing.

@taskinkemal please add a check for `IsAbstract` method:
```suggestion
			return method.IsVirtual || method.IsOverride || method.IsAbstract;
```

taskinkemal: Sure, done. Thanks.

dropsonic (approved):

docs/Summary.md
EAndrosova: approved

docs/diagnostics/PX1096.md
EAndrosova: Please replace the diagnostic description with the following: 

# PX1096
This document describes the PX1096 diagnostic.

## Summary

| Code   | Short Description                                                                     | Type  | Code Fix    | 
| ------ | ------------------------------------------------------------------------------------- | ----- | ----------- | 
| PX1096 |Signature of a method with the `PXOverride` attribute must match the overriden method. | Error | Unavailable | 

## Diagnostic Description
Signature of a method with the `PXOverride` attribute must match the overriden method.
The signatures of the overriden and overriding methods must comply with one of the following rules:
* Signatures of both methods are identical.
* Signatures of both methods are identical, except the overriding method has one additional parameter, which is a delegate that points to the base method. In this scenario, the delegate signature must also match the overriden method.

## Further Constraints
* The overriding method cannot be static
* The overriding method cannot be generic
* The base method must be a member of graph extension, that is a class that derives from `PXGraphExtension`
* The base method must be virtual (have either the virtual or the override signature)
* The base method must have one of the following accessibility levels: public, protected or protected internal
* The names of the derived method and the base method must match

## Example of Correct Code

```C#
public class MyGraph : PXGraph<MyGraph>
public class BaseGraphExtension : PXGraphExtension<MyGraph>
{
	public virtual int Add(int x, string y)
	{
		return x + Convert.ToInt32(y);
	}
}
public class DerivedGraphExtension : PXGraphExtension<BaseGraphExtension, MyGraph>
{
	[PXOverride]
	public virtual int Add(int x, string y)
	{
		return x + Convert.ToInt32(y) * 2;
	}
}
```

## Example of Correct Code with Delegate

```C#
public class MyGraph : PXGraph<MyGraph>
public class BaseGraphExtension : PXGraphExtension<MyGraph>
{
	public virtual int Add(int x, string y)
	{
		return x + Convert.ToInt32(y);
	}
}
public class DerivedGraphExtension : PXGraphExtension<BaseGraphExtension, MyGraph>
{
	[PXOverride]
	public virtual int Add(int x, string y, Func<int, string, int> del)
	{
		if (x < 10)
		{
			return x + Convert.ToInt32(y) * 2;
		}
		
		return del(x, y);
	}
}
```

## Example of Incorrect Code
In the following example, the delegate signature does not correspond to the signature of the base method.
```C#
public class MyGraph : PXGraph<MyGraph>
public class BaseGraphExtension : PXGraphExtension<MyGraph>
{
	public virtual int Add(int x, string y, bool z)
	{
		return x + Convert.ToInt32(y);
	}
}
public class DerivedGraphExtension : PXGraphExtension<BaseGraphExtension, MyGraph>
{
	[PXOverride]
	public virtual int Add(int x, string y, bool z, Func<int, string, int> del)
	{
		if (x < 10 && z)
		{
			return x + Convert.ToInt32(y) * 2;
		}
		
		return del(x, y);
	}
}
```

## Related Articles
[To override a Virtual Method](https://help.acumatica.com/(W(16))/Help?ScreenId=ShowWiki&pageid=6fa2a444-17b4-42f9-9e6a-64e85167626a)
[PXOverrideAttribute](https://help.acumatica.com/Help?ScreenId=ShowWiki&pageid=4bd32f02-a673-3a01-1748-33a54f3cd035)

src/Acuminator/Acuminator.Analyzers/Resources.resx
EAndrosova: Please replace with the following:
The signature of the method with the PXOverride attribute must match the overriden method

EAndrosova (commented):
please update the diagnostic description according to comments

EAndrosova (commented):
please update the diagnostic description according to comments

docs/diagnostics/PX1096.md
taskinkemal: Updated. Thank you.

src/Acuminator/Acuminator.Analyzers/Resources.resx
taskinkemal: Updated. Thank you.

SENya1990:
@taskinkemal many thanks for your contribution. New diagnostic will be shipped with the new Acuminator version

dropsonic:
@taskinkemal thank you so much for contributing!

PR #516, opened 2024-02-16
Bugfix/atr 792 dev update to csharp 12 version

SENya1990:
**Merge after https://github.com/Acumatica/Acuminator/pull/515**

Changes Overview:
- increased C# version to 12 (initial goal was 11 but there is no reason to postpone using C# 12 and new collection creation expression)
- Replaced old Nullable polyfill package with PolySharp.
- Added `System.Memory` package for `Span` types.
- Investigated possibility for more polyfills. Unfortunately, the addition of polyfill for range operators for array is too difficult and unstable because it requires to add a member to already existing type `System.Runtime.CompilerServices.RuntimeHelpers`


src/Acuminator/Acuminator.Analyzers/Acuminator.Analyzers.csproj
```diff
@@ -24,6 +24,11 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
+    <PackageReference Include="PolySharp" Version="1.14.1">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+    </PackageReference>
+    <PackageReference Include="System.Memory" Version="4.5.5" />
```
dropsonic: Can `System.Memory` lead to DLL hell in Visual Studio (VSIX) or Rider (NuGet package)?
AFAIK, adding external dependencies to Roslyn analyzers is not recommended.

SENya1990: Well, it works in my VS. Not a good argument, I know. And we had such problems previously with value tuples nuget package. 

Imho, it's still worth to try it:
-  It works on the latest VS 2022 on my machine in the scenario where value tuples didn't. It may break in theory on some older VS version. But we are not really willing to support them.
- We already have dependencies in analyzers. The whole Roslyn is a dependency actually. Plus we have some additional dependencies in VSIX and Utils: 
https://github.com/Acumatica/Acuminator/blob/ef8e53c16cdd94885ae26203280b7e5f3aef7b3a/src/Acuminator/Acuminator.Utilities/Acuminator.Utilities.csproj#L29
- Spans are important new .Net APIs that are supported even by the language. Excluding them decreases the scope of features that we may use from the latest C# versions. For example, we already can't make a polyfill for collection expressions `[..]` from C# 12 that will work with arrays because the compiler requires that an existing framework type declares a special method that it does not have in old .Net.  The workaround is to use spans that are supported:
```cs
var span = array.AsSpan();
return span[1..];
```

dropsonic: Acuminator is used as a NuGet package in Rider / VS Code / build tools, so the fact that it works in the VS does not prove anything. Roslyn is an expected dependency that is provided by the runtime that calls the analyzer.
VSIX is a separate thing; I'm talking about the analyzer package only.

Spans are important, but they are not required. Everything perfectly works without them. I see a lot of risks here and do not see a clear benefit that worth this risk.

SENya1990: Well, in that case you are probably right. We can live without them, I don't have any scenarios in my head that will be drastically improved by spans. If I have one we can reconsider, but for now let's remove this dependency from analyzers and utils.

SENya1990: done

src/Acuminator/Acuminator.Vsix/Properties/AssemblyInfo.cs
```diff
@@ -5,9 +5,9 @@
 [assembly: AssemblyConfiguration("")]
 [assembly: AssemblyCompany("")]
 [assembly: AssemblyProduct("Acuminator")]
-[assembly: AssemblyCopyright("Copyright � 2017-2023 Acumatica Ltd.")]
+[assembly: AssemblyCopyright("Copyright � 2017-2024 Acumatica Ltd.")]
```
dropsonic: I think we should consider moving it to the `csproj` file. It allows us to use a dynamic value instead of manually changing it every year. By the way, it is not `Acumatica Ltd.` anymore, it is `Acumatica, Inc.`
An example:

```xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <Company>Acumatica, Inc.</Company>
    <Authors>$(Company)</Authors>
    <Copyright>Copyright © 2017-$([System.DateTime]::Now.Year) $(Company)</Copyright>
    ...
  </PropertyGroup>
  ...
</Project>
```

SENya1990: It would be great to have no need to update copyright. Unfortunately it just does not work with the old non-SDK project format. So, I have updated the Analyzers and Utilities projects where it works. Also I have fixed info about company in VSIX and tests projects.

dropsonic (approved):

dropsonic (approved):

PR #517, opened 2024-02-16
Bugfix/atr 789 dev change acumatica logo

SENya1990:
**Merge after https://github.com/Acumatica/Acuminator/pull/516**

Overview:
- Updated Acumatica logo in all places.


SENya1990:
@dropsonic maybe we should not update the logo for icons in VS menu? IMHO it looks a bit better with the old logo because it is simpler and the icon is small. Compare this:

_New logo:_
image

_Old logo:_
image

src/Acuminator/Acuminator.Analyzers/packageIcon.png
dropsonic, line 1: The original one was `256x256`, the new one is not even rectangular, it is `2000x1973`. Is it OK with NuGet? I suppose we should follow its recommendations.

SENya1990: Yes, we probably should do it. Do we need to ask the creator of the icon for the 256x256 resolution, or some online resizer will be ok?

dropsonic: Yes, please write to Drew Donaghy and request an icon.

dropsonic: Yes, please write to Drew Donaghy and request an icon.

SENya1990: Received an icon from Drew Donaghy and replaced this icon with it.

dropsonic:
> @dropsonic maybe we should not update the logo for icons in VS menu? IMHO it looks a bit better with the old logo because it is simpler and the icon is small. Compare this:
> 
> _New logo:_ image
> 
> _Old logo:_ image

Please consult with Drew on that.

SENya1990:
@dropsonic got a reply that we should use the new logo everywhere. So, we can't keep the old logo.

dropsonic (approved):

PR #543, opened 2024-08-20
Bugfix/atr 821 support infos from metadata

SENya1990:
Made significant changes to Acuminator code base to enhance semantic models with the collection of data from the base types that can be declared in external assemblies. We didn't have it previously and it greatly matters for the analysis of customizations.

These also are main analysis infrastructure changes to support enhanced DAC analysis features:
- check DAC fields for the presence of BQL fields including base types
- check DAC fields to check for the presence of timestamp and audit fields
- display in Code Map nodes representing base code elements - base DACs, base Graphs, base DAC fields, base graph Views/Actions, etc

In addition, to better check that new changes doesn't introduce new NREs nullable code analysis was added to Analyzers, Utilities and Vsix projects. Added nullable annotations and checks wherever necessary.

Other changes - refactoring and code enhancements.


dropsonic:
This PR contains 378 changed files, mostly with whitespaces/line breaks — what are the real changes?
Can the functional and style changes be separated?

SENya1990:
@dropsonic  the real changes are described in the PR summary:
https://github.com/Acumatica/Acuminator/pull/543#issue-2476597782

In most of the files there are mostly the nullability analysis changes.
Since the nullable analysis was added to three projects the changes are big but they usually are put into separate commits which I believe are more granular.

The nullable analysis was enabled on such a big scope because now the `Node` in semantic infos may be null and we need to ensure that any place calling it does not throw NRE.
And there are also removed redundant namespaces, and renamed old analyzer because it had an incorrect name.

I will try to make PRs more granular in the future but it would require more work from me. Frankly, while I am the only one adding the code, I feel this is an overkill. That is also the reason why I can freely add such changes. If there is someone else working on Acuminator, then definitely the changes should be more granular.

dropsonic:
I do not mind

dropsonic (approved):

PR #559, opened 2024-09-22
Bugfix/atr 829 dev rework px1067 to report and redeclare all fields at once

SENya1990:
**Merge after https://github.com/Acumatica/Acuminator/pull/558**

**Changes Overview**
- Reworked PX1067 to report all not re-declared BQL fields at one in a single warning due to a huge number of errors on a single node. 
- PX1067 now always reports warning on the DAC name
- Changed diagnostic messages to support case with multiple not re-declared BQL fields
- Reworked PX1067 code fix. Now it will re-declare all reported BQL fields at once and supports batch fixing
- Extended unit tests


src/Acuminator/Acuminator.Analyzers/Resources.resx
```diff
@@ -609,10 +609,28 @@ public virtual int? SomeDacField{ get; set; }</value>
   <data name="PX1066TitleFormat" xml:space="preserve">
     <value>A possible typo in the BQL field name. The DAC field property name for this BQL field: {0}.</value>
   </data>
-  <data name="PX1067FixFormat" xml:space="preserve">
+  <data name="PX1067From_2_To_5_FieldsFixFormat" xml:space="preserve">
```
SENya1990, line 612: @EAndrosova  please review changes in the PX1067 diagnostic and code fix messages. 
Since the diagnostic now reports all not re-declared BQL fields at once, there are now three strings for three cases:
- Only one BQL field is reported
- From 2 to 5 BQL fields are reported
- More than 5 BQL fields are reported

dropsonic (approved):

src/Acuminator/Acuminator.Analyzers/Resources.resx
EAndrosova: Could you please change the strings as follows:
Redeclare the {0} BQL fields in the "{1}" DAC - okay

BQL fields from a base DAC should be redeclared in the derived DAC. The "{0}" DAC does not contain a redeclaration of {1} BQL fields. >> BQL fields from the base DAC should be redeclared in the derived DAC. The "{0}" DAC does not contain a redeclaration of {1} BQL fields.

Redeclare the {0} and {1} more BQL {2} in the "{3}" DAC - could you please clarify what is stands for 0, 1, and 2?

SENya1990: @EAndrosova {0} stands for names of five BQL fields joined with a comma like this:
```
"lineNbr", "docType", "refNbr", "tstamp", "docDate"
```

{1} stands for the number of remaining reported BQL fields that didn't fit into the message and were cut off. 
For instance, if we have 8 reported fields, then names of 5 BQL fields will be displayed in {0} and the {1} will be `3`

{2} - depends on the count of remaining BQL fields in {1}. If there is only 1 remaining field then {2} is `field`, if there is more than 1 remaining field, then {2} is `fields`.

For example:
```
Redeclare the "lineNbr", "docType", "refNbr", "tstamp", "docDate" and 1 more BQL field in the "{3}" DAC

Redeclare the "lineNbr", "docType", "refNbr", "tstamp", "docDate" and 2 more BQL fields in the "{3}" DAC
```

PR #561, opened 2024-09-23
Updated PX1099.md documentation and fixed mistypes

SENya1990:


docs/diagnostics/PX1099.md
```diff
@@ -54,8 +54,8 @@ public class MyGraph : PXGraph<MyGraph>
 The main idea of the PX1099 analysis is that all API usages in the checked code are checked against a list of forbidden APIs. The API is reported if one of the following criteria is met:
 
 - The API itself is in the list of banned APIs.
-- A namespace contains the API that is in the list of banned APIs.
-- One of the types contains the API that is is in the list of banned APIs.
+- A namespace containing the API is in the list of banned APIs.
+- The type containing the API is in the list of banned APIs. In case of nested types, the API is banned if any of containing types is in the list of banned APIs.
```
EAndrosova, line 58: maybe change to the following to make is a little bit more clear? or is it okay as it is?
In case of nested types, the API is banned if any of the types that contain the API is in the list of banned APIs.

EAndrosova (approved):

docs/diagnostics/PX1099.md
SENya1990: I think, it's clear enough like this.

PR #573, opened 2024-11-05
Updated how Acumatica version is determined to be more durable in order to resolve an issue where PX1042 was improperly reported in Acumatica 2023 R1 and up.  This was caused by Acuminator determining whether or not the Acumatica version was above a certain range by checking for the existence of Types/Members, one being PXCache.RowSelectingWhileReading which was internal, which was not returned by Roslyn because it is marked as internal.

### LINKED ISSUE #572: PX1042_DatabaseQueriesInRowSelecting shows error even in 2023R1 and up
Stepping through the code when running the Unit Tests it seems that in the constructor of `Acuminator.Utilities.Roslyn.Semantic.Symbols.PXCacheSymbols`, `RowSelectingWhileReading` is always null after :

```cs
Type.GetMembers(EventsNames.PXCache.RowSelectingWhileReading)
.OfType<IEventSymbol>()
.FirstOrDefault();
```

because `RowSelectingWhileReading` is internal, and internal members aren't returned by `GetMembers`.

I was able to alter `Acuminator.Tests.Verification.VerificationHelper.CreateProject` to have this :

```cs
CSharpCompilationOptions compileOptions = new CSharpCompilationOptions(OutputKind.DynamicallyLinkedLibrary)
.WithMetadataImportOptions(MetadataImportOptions.All); // compile to a dll
```

However, I'm not sure if there's a way to do that for the analyzer when it's part of the installed extension.

I was also unable to step through the code when debugging an instance of VS, because for some reason the methods weren't getting hit specifically for `DatabaseQueriesInRowSelectingAnalyzer`. I've been able to get other analyzers to hit breakpoints while debugging an instance of VS.

ShadowFiendZX:
Updated how Acumatica version is determined to be more durable in order to resolve an issue where PX1042 was improperly reported in Acumatica 2023 R1 and up.  This was caused by Acuminator determining whether or not the Acumatica version was above a certain range by checking for the existence of Types/Members, one being PXCache.RowSelectingWhileReading which was internal, which was not returned by Roslyn because it is marked as internal.

Also fixed a test which was not passing due to a "compilation" error that was not reported by the test runner, but was accessible while debugging.

### DIFF
--- a/src/Acuminator/Acuminator.Tests/Tests/StaticAnalysis/ChangesInPXCache/ChangesInPXCacheDuringPXGraphInitializationTests.cs
+++ b/src/Acuminator/Acuminator.Tests/Tests/StaticAnalysis/ChangesInPXCache/ChangesInPXCacheDuringPXGraphInitializationTests.cs
@@ -9,52 +9,52 @@ using Xunit;
 
 namespace Acuminator.Tests.Tests.StaticAnalysis.ChangesInPXCache
 {
-    public class ChangesInPXCacheDuringPXGraphInitializationTests : DiagnosticVerifier
-    {
-        protected override DiagnosticAnalyzer GetCSharpDiagnosticAnalyzer() =>
-            new PXGraphAnalyzer(
-                CodeAnalysisSettings.Default
-                .WithRecursiveAnalysisEnabled(),
-                new ChangesInPXCacheDuringPXGraphInitializationAnalyzer());
+	public class ChangesInPXCacheDuringPXGraphInitializationTests : DiagnosticVerifier
+	{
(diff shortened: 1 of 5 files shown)


SENya1990:
@ShadowFiendZX  this way of checking is not correct. As I mentioned in the issue, `AssemblyFileVersionAttribute` can't be used for such checks because it is set only in Acumatica production builds. Acumatica dlls compiled locally on developer machines do not have it:
image

Moreover, the fix requires more than jut a change of check for PX1042 analyzer. The fix requires Roslyn to start loading all metadata. In this case I plan to work on the fix personally.

PR #592, opened 2025-07-03
Bugfix/atr 633 dev px1031 allow to string

SENya1990:
**Changes Overview:**
- enhanced PX1031 analysis for instance methods in DACs and DAC extensions by allowing `ToString` overrides
- added `IsOverride` syntax node helper
- added scenario with `ToString` to PX1031 unit tests
- extended documentation for PX1031

### DIFF
--- a/docs/diagnostics/PX1031.md
+++ b/docs/diagnostics/PX1031.md
@@ -9,7 +9,7 @@ This document describes the PX1031 diagnostic.
 
 ## Diagnostic Description
 DACs and DAC extensions cannot contain instance methods. DACs and DAC extensions are the classes that are used to read and write data. Therefore, for appropriate program architecture and design, such classes cannot contain any application logic.
-The only exception to this rule is an override of the `ToString` method which can be useful to provide DAC text representation.
+The only exception to this rule is an override of the `ToString` method which can be useful to provide the string representation of the DAC and its fields.
 
 To fix the issue, you should remove instance methods from the DAC or the DAC extension.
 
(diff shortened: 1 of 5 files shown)


docs/diagnostics/PX1031.md
```diff
@@ -8,7 +8,8 @@ This document describes the PX1031 diagnostic.
 | PX1031 | DACs cannot contain instance methods. | Error | Unavailable | 
 
 ## Diagnostic Description
-DACs and DAC extensions cannot contain instance methods. DACs and DAC extensions are the classes that are used to read and write data. Therefore, for appropriate program architecture and design, such classes cannot contain any application logic. 
+DACs and DAC extensions cannot contain instance methods. DACs and DAC extensions are the classes that are used to read and write data. Therefore, for appropriate program architecture and design, such classes cannot contain any application logic.
```
SENya1990, line 11: @EAndrosova  please check changes in the documentation.

docs/diagnostics/PX1031.md
```diff
@@ -8,7 +8,8 @@ This document describes the PX1031 diagnostic.
...
+The only exception to this rule is an override of the `ToString` method which can be useful to provide DAC text representation.
```
dropsonic: In theory, [`Deconstruct`](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/functional/deconstruct#user-defined-types) is another valid case of having an instance method in DAC, but I don't expect it to be very popular (and it can also be done via a static extension method outside of DAC class)

SENya1990: Let's add it if there actually is a request for it. There is a workaround + I haven't seen its usage in DACs yet. There was a request for `ToString`, any other theoretical scenarios may just eat performance without a benefit.

EAndrosova (approved):

dropsonic (approved):

PR #663, opened 2026-03-10
ATR-955: Added severity to the output

weihnachtswichtel:
### Changes Overview
- Extend `Line` to support 4 spans (severity + id + message + location) and update `ToString()` for 4-span lines.
- Update grouping to populate severity when producing diagnostic report `Line`s.
- Update plain-text file formatting to print severity with the diagnostic line.

### DIFF
--- a/src/Acuminator/Acuminator.Runner.NetFramework/Output/Grouping/Base/GroupLinesBase.cs
+++ b/src/Acuminator/Acuminator.Runner.NetFramework/Output/Grouping/Base/GroupLinesBase.cs
@@ -55,7 +55,7 @@ namespace Acuminator.Runner.Output.Grouping
 																		   Location: GetPrettyLocation(d, projectDirectory, analysisContext)));
 
 			var sortedDiagnosticInfos = GetSortedDiagnosticInfos(sortByDiagnosticId, sortBySourceFile, unsortedDiagnosticInfos);
-			var reportLines	= sortedDiagnosticInfos.Select(d  => new Line(d.Diagnostic.Id, d.Message, d.Location));
+			var reportLines	= sortedDiagnosticInfos.Select(d  => new Line(d.Diagnostic.Severity.ToString(), d.Diagnostic.Id, d.Message, d.Location));
 
 			return reportLines;
 		}
--- a/src/Acuminator/Acuminator.Runner.NetFramework/Output/Json/Converters/LineConverter.cs
+++ b/src/Acuminator/Acuminator.Runner.NetFramework/Output/Json/Converters/LineConverter.cs
@@ -14,6 +14,8 @@ namespace Acuminator.Runner.Output.Json
 	/// </summary>
 	internal class LineConverter : JsonConverter<Line>
 	{
+		protected const string LinePartsSeparator = ": ";
+		protected const string SeverityTemplate = "[{0}] ";
 		public override Line Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options) =>
 			throw new NotSupportedException();
 
@@ -35,7 +37,15 @@ namespace Acuminator.Runner.Output.Json
 					{
 						var (diagnosticId, diagnosticMessage, location) = (line.Spans[0].ToString(), line.Spans[1].ToString(), line.Spans[2].ToString());
 
-						writer.WriteStringValue($"{diagnosticId}: {diagnosticMessage}: {location}");
+						writer.WriteStringValue($"{diagnosticId}{LinePartsSeparator}{diagnosticMessage}{LinePartsSeparator}{location}");
+						return;
+					}
+
+				case 4:
+					{
+						var (severity, diagnosticId, diagnosticMessage, location) = (line.Spans[0].ToString(), line.Spans[1].ToString(), line.Spans[2].ToString(), line.Spans[3].ToString());
+
+						writer.WriteStringValue($"{string.Format(SeverityTemplate,severity)}{diagnosticId}{LinePartsSeparator}{diagnosticMessage}{LinePartsSeparator}{location}");
 						return;
 					}
 
--- a/src/Acuminator/Acuminator.Runner.NetFramework/Output/PlainText/PlainTextReportOutputterConsole.cs
+++ b/src/Acuminator/Acuminator.Runner.NetFramework/Output/PlainText/PlainTextReportOutputterConsole.cs
@@ -60,7 +60,12 @@ namespace Acuminator.Runner.Output.PlainText
 												diagnosticMessage: line.Spans[1].ToString(),
 												location: line.Spans[2].ToString());
 					return;
-					
+				case 4:
+					WriteDiagnosticWithLocation(indentationLevel, diagnosticId: line.Spans[1].ToString(),
+												diagnosticMessage: line.Spans[2].ToString(),
+												location: line.Spans[3].ToString(),
+												severity: line.Spans[0].ToString());
+					return;
 				case 1:
 				default:
 					string padding = GetPadding(indentationLevel);
@@ -69,12 +74,18 @@ namespace Acuminator.Runner.Output.PlainText
 			}
 		}
 
-		private void WriteDiagnosticWithLocation(int indentationLevel, string diagnosticId, string diagnosticMessage, string location)
+		private void WriteDiagnosticWithLocation(int indentationLevel, string diagnosticId, string diagnosticMessage, string location, string? severity = null)
 		{ 
 			string padding = GetPadding(indentationLevel);
 			var oldColor   = Console.ForegroundColor;
 
-			WriteStringPartWithColor(padding + diagnosticId, ConsoleColor.White);
+			Console.Write(padding);
+			if (!string.IsNullOrEmpty(severity))
+			{
+				WriteStringPartWithColor(String.Format(SeverityTemplate, severity), GetSeverityColor(severity!));
+			}
+
+			WriteStringPartWithColor(diagnosticId, ConsoleColor.White);
 			Console.Write(LinePartsSeparator);
 			Console.Write(diagnosticMessage + LinePartsSeparator);
 			WriteStringPartWithColor(location, ConsoleColor.Yellow);
@@ -92,6 +103,21 @@ namespace Acuminator.Runner.Output.PlainText
 				finally
 				{
 					Console.ForegroundColor = oldColor;
+				}
+			}
+
+			static ConsoleColor GetSeverityColor(string severity)
+			{
+				switch(severity)
+				{
+					case SeverityError:
+						return ConsoleColor.Red;
+					case SeverityWarning:
+						return ConsoleColor.Yellow;
+					case SeverityInfo:
+						return ConsoleColor.Cyan;
+					default:
+						return Console.ForegroundColor;
 				}
 			}
 		}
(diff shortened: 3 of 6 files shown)


src/Acuminator/Acuminator.Runner.NetFramework/Output/PlainText/PlainTextReportOutputterFile.cs
```diff
@@ -88,9 +88,9 @@ protected override void WriteLine(in Line line, int indentationLevel)
 
 			if (line.Spans.Length == 3)
```
SENya1990, line 88: Yes, here is a bug. I would suggest either changing this to 4, or keep two cases for 3 and 4 spans.

src/Acuminator/Acuminator.Runner.NetFramework/Output/PlainText/PlainTextReportOutputterFile.cs
```diff
@@ -88,9 +88,9 @@ protected override void WriteLine(in Line line, int indentationLevel)
...
 			{
-				var (diagnosticId, diagnosticMessage, location) = 
-					(line.Spans[0].ToString(), line.Spans[1].ToString(), line.Spans[2].ToString());
-				WriteLine($"{padding}{diagnosticId}{LinePartsSeparator}{diagnosticMessage}{LinePartsSeparator}{location}");
+				var (severity, diagnosticId, diagnosticMessage, location) = 
+					(line.Spans[0].ToString(), line.Spans[1].ToString(), line.Spans[2].ToString(), line.Spans[3].ToString());
+				WriteLine($"{padding}{severity} - {diagnosticId}{LinePartsSeparator}{diagnosticMessage}{LinePartsSeparator}{location}");
```
SENya1990: I'm also not sure about the style here. Maybe we should have it as something like this:
`[Warning] PX1XXX:<message>:<location>`
It looks closer to what we had in RVT.

Also separator string for Severity should be made a constant similar to `LinePartsSeparator`

SENya1990:
Severity is not supported for JSON output  (see `LineConverter` class)

src/Acuminator/Acuminator.Runner.NetFramework/Output/Grouping/Base/GroupLinesBase.cs
```diff
@@ -55,7 +55,7 @@ protected IEnumerable<Line> GetOrderedDiagnosticsWithLocationLines(IEnumerable<D
 																		   Location: GetPrettyLocation(d, projectDirectory, analysisContext)));
 
 			var sortedDiagnosticInfos = GetSortedDiagnosticInfos(sortByDiagnosticId, sortBySourceFile, unsortedDiagnosticInfos);
-			var reportLines	= sortedDiagnosticInfos.Select(d  => new Line(d.Diagnostic.Id, d.Message, d.Location));
+			var reportLines	= sortedDiagnosticInfos.Select(d  => new Line(d.Diagnostic.Severity.ToString(), d.Diagnostic.Id, d.Message, d.Location));
```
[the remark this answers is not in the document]
SENya1990: @weihnachtswichtel , AI is right here, there is no support for `PlainTextReportOutputterConsole` and  `LineConverter`, and that should be fixed. It's even more work for the `PlainTextReportOutputterConsole` since it outputs text to console in colors and I would suggest to use different colors for different severities.

src/Acuminator/Acuminator.Runner.NetFramework/Output/Grouping/Base/GroupLinesBase.cs
SENya1990, line 71: Perhaps, we need to order diagnostics by severity before ordering them by ID, but I'm not sure about this. It's up to discussion and I won't insist on it if it doesn't look useful

SENya1990 (changes requested):
There are several unaddressed cases now, and even a bug with accessing array element out of array bounds, the fix can't be merged as is

weihnachtswichtel:
Reported items resolved

src/Acuminator/Acuminator.Runner.NetFramework/Output/PlainText/PlainTextReportOutputterConsole.cs
```diff
@@ -69,12 +74,18 @@ protected override void WriteLine(in Line line, int indentationLevel)
...
 
-		private void WriteDiagnosticWithLocation(int indentationLevel, string diagnosticId, string diagnosticMessage, string location)
+		private void WriteDiagnosticWithLocation(int indentationLevel, string diagnosticId, string diagnosticMessage, string location, string? severity = null)
 		{ 
 			string padding = GetPadding(indentationLevel);
 			var oldColor   = Console.ForegroundColor;
 
-			WriteStringPartWithColor(padding + diagnosticId, ConsoleColor.White);
+			Console.Write(padding);
+			if (!string.IsNullOrEmpty(severity))
```
SENya1990: Please use the extension method `IsNullOrEmpty` over the regular `string.IsNullOrEmpty`. The regular `string.IsNullOrEmpty` method on .Net Framework (and this runner targets .Net Framework) does not have nullable annotations. It was added to this method in newer .Net runtimes. On the other hand, the extension method `IsNullOrEmpty` defined in Acuminator has proper annotations for the C# nullable analysis.

src/Acuminator/Acuminator.Runner.NetFramework/Output/PlainText/PlainTextReportOutputterConsole.cs
```diff
@@ -69,12 +74,18 @@ protected override void WriteLine(in Line line, int indentationLevel)
...
+			{
+				WriteStringPartWithColor(String.Format(SeverityTemplate, severity), GetSeverityColor(severity!));
```
SENya1990: If you implement the comment https://github.com/Acumatica/Acuminator/pull/663/changes#r2917541888, you can remove the nullability erasure operator and re;y on nullable analysis

```suggestion
				WriteStringPartWithColor(String.Format(SeverityTemplate, severity), GetSeverityColor(severity));
```

src/Acuminator/Acuminator.Runner.NetFramework/Output/Json/Converters/LineConverter.cs
```diff
@@ -14,6 +14,8 @@ namespace Acuminator.Runner.Output.Json
 	/// </summary>
 	internal class LineConverter : JsonConverter<Line>
 	{
+		protected const string LinePartsSeparator = ": ";
```
SENya1990, line 17: I have a feeling that we need to create a separate file with constants since there are copies for some of them here and in `PlainTextReportOutputterBase`. 

Not insisting on doing it in this PR, I can think about where to put them later.

src/Acuminator/Acuminator.Runner.NetFramework/Output/PlainText/PlainTextReportOutputterConsole.cs
```diff
@@ -69,12 +74,18 @@ protected override void WriteLine(in Line line, int indentationLevel)
 			}
 		}
 
-		private void WriteDiagnosticWithLocation(int indentationLevel, string diagnosticId, string diagnosticMessage, string location)
+		private void WriteDiagnosticWithLocation(int indentationLevel, string diagnosticId, string diagnosticMessage, string location, string? severity = null)
```
SENya1990: Not insisting, but I wouldn't use optional parameter here. The optional parameter won't bring any convenience benefits here because it's a private method and it is called only twice and only 1 time without a severity. 

IMHO, in this case it just loses the clarity of the calling code with implicitly inserted optional value and it's better to explicitly pass `severity: null`.

src/Acuminator/Acuminator.Runner.NetFramework/Output/PlainText/PlainTextReportOutputterConsole.cs
```diff
@@ -94,6 +105,21 @@ static void WriteStringPartWithColor(string part, ConsoleColor color)
 					Console.ForegroundColor = oldColor;
 				}
 			}
+
+			static ConsoleColor GetSeverityColor(string severity)
+			{
+				switch(severity)
```
SENya1990: This looks like a good place to use switch expression:
```cs
static ConsoleColor GetSeverityColor(string severity) => severity switch
{
    SeverityError   => ConsoleColor.Red,
    SeverityWarning => ConsoleColor.Yellow,
    SeverityInfo    => ConsoleColor.Cyan,
    _               => Console.ForegroundColor
};
```
Not insisting since this is about code style.

SENya1990: Another note is that this code has an external dependency that the severity will always be in proper casing and format. This won't work on "warning" or "error".

Currently, it should be ok, since it is just a result of the ToString call on enum, but you can't be sure it will always be like that. It would be more robust to have case insensitive comparisons with one of the following:

- Use constants in upper casing and convert severity to upper case before checking
- Use static readonly dictionary with case insensitive comparer for mapping between severity and console color. This is the most flexible option
- Use direct string equals calls with case insensitive comparison.

It doesn't really matter which option will be selected.

SENya1990: In general, I think the design should be somehow re-though in the future - the fix is looking simple, but I feel that there is an issue with the separation of concepts. The output of data is late in pipeline and it works with abstract report data model - groups, lines, spans, etc.

It's wrong to put business logic and terms there. It shouldn't know anything about the severity. 
I don't think that this needs to be addressed in this PR, I will think about it later.

src/Acuminator/Acuminator.Runner.NetFramework/Output/PlainText/PlainTextReportOutputterConsole.cs
```diff
@@ -69,12 +74,18 @@ protected override void WriteLine(in Line line, int indentationLevel)
 			}
 		}
 
-		private void WriteDiagnosticWithLocation(int indentationLevel, string diagnosticId, string diagnosticMessage, string location)
+		private void WriteDiagnosticWithLocation(int indentationLevel, string diagnosticId, string diagnosticMessage, string location, string? severity = null)
 		{ 
 			string padding = GetPadding(indentationLevel);
 			var oldColor   = Console.ForegroundColor;
```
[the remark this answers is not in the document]
SENya1990: Seems like it should be enough to remove the variable.

SENya1990 (changes requested):
@weihnachtswichtel the bugs were resolved. The new remarks can be separated into two groups:
- The small remarks about code style and code warnings. There are some small issues with handling C# nullability analysis. Code can be more robust in some places. AI has sensible remarks. Some of the remarks are marked as optional via "Not insisting" sentence. This group should be addressed.
- The general remarks about design issues that IMO shouldn't be addressed in scope of this PR.

src/Acuminator/Acuminator.Runner.NetFramework/Output/PlainText/PlainTextReportOutputterConsole.cs
weihnachtswichtel: Agree, that is fixed

src/Acuminator/Acuminator.Runner.NetFramework/Output/PlainText/PlainTextReportOutputterConsole.cs
weihnachtswichtel: Used first option

src/Acuminator/Acuminator.Runner.NetFramework/Output/PlainText/PlainTextReportOutputterConsole.cs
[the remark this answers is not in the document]
weihnachtswichtel: unused variable removed

weihnachtswichtel:
@SENya1990, issues resolved, please review

SENya1990 (approved):

PR #668, opened 2026-05-06
Significant performance improvement

gelverpl:
### Proposed changes

1. `NestedInvocationWalker` now caches retrieved `ISymbol` instances using a new `SymbolInfoCache`.
2. Fixed a serious issue that was causing a large amount of redundant work: multiple call graph traversals, which in turn caused an excessive number of semantic queries to resolve `ISymbol`.

### Motivation

While profiling the analyzer, I noticed extremely high memory usage, and running solution analysis from the CLI took more than 40 minutes.
dotTrace pointed to `Acuminator.Utilities.Roslyn.NestedInvocationWalker.GetSymbol(ExpressionSyntax)` as a primary hotspot:
Screenshot 2026-05-06 201213
Screenshot 2026-05-06 201339

Memory profiling confirmed the same method:
Screenshot 2026-05-06 204141

### Caching

As a load-reducing measure, I implemented caching. This significantly reduced both execution time and memory consumption.

Yes, caching is often a last-resort approach and ideally **shouldn't be** here. However, when implemented carefully it can also provide valuable diagnostic data that helps guide further optimization work.

### Cache statistics

`SymbolInfoCache` can optionally collect statistics. Enable it by defining the `SYMBOL_INFO_CACHE_STATISTICS` constant, then inspect `Acuminator.Utilities.Roslyn.SymbolInfoCache.SymbolInfoCacheStatisticsContext.GetStatistics` under the debugger.

I intentionally did not add any logging of it because the statistics can become very large. It's much more practical to analyze it interactively (e.g., via LINQ in a debug session).

While digging into the stats, I noticed something weird: the same expression was cached under two different key types: `InvocationExpressionSyntax` and `MemberAccessExpressionSyntax`. Debugging led to `VisitMemberAccessExpression`.

### `foo.Bar()` is both `MemberAccessExpressionSyntax` and `InvocationExpressionSyntax`

By filtering out redundant invocation processing inside `MemberAccessExpressionSyntax`, I added an additional fix. This change, similarly to caching, delivered a substantial improvement in both time and memory.

### Benchmarks

I measured the impact of each fix independently, as well as both together.

**Total runtime (wall-clock):**
Screenshot 2026-05-06 211607

- Speedup ranges from **1.30x to 7.43x** (about **4x on average**)

**Memory allocation / GC pressure:**
Screenshot 2026-05-06 211648

- Allocations reduced by **1.59x to 8.78x** (about **5x on average**)
- Significantly less GC overhead due to drastically fewer allocations

With both fixes applied, traces no longer show `GetSymbol` as a top hotspot; instead, the cache path (`GetOrCreate`) appears, and overall memory usage is dramatically improved.
Screenshot 2026-05-06 203630
Screenshot 2026-05-06 204713

### Conclusion

Even though caching contributes less after the second fix, I'd still like to keep it (at least for upcoming optimization work) as a way to collect detailed statistics. The goal is to reach a hit rate that is **0 or as close to 0 as possible** (i.e., eliminate the underlying redundant symbol queries altogether).

I plan to continue working on this area and believe we can achieve even more significant improvements.

### DIFF
--- a/src/.idea/.idea.Acuminator/.idea/vcs.xml
+++ b/src/.idea/.idea.Acuminator/.idea/vcs.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="VcsDirectoryMappings">
+    <mapping directory="$PROJECT_DIR$/.." vcs="Git" />
+  </component>
+</project>
--- a/src/Acuminator/Acuminator.Utilities/Roslyn/NestedInvocationWalker.cs
+++ b/src/Acuminator/Acuminator.Utilities/Roslyn/NestedInvocationWalker.cs
@@ -210,8 +210,16 @@ namespace Acuminator.Utilities.Roslyn
 
 		public override void VisitMemberAccessExpression(MemberAccessExpressionSyntax node)
 		{
-			VisitPropertyOrIndexerAccessExpression(node);
-			base.VisitMemberAccessExpression(node);
+			if (node.Parent is InvocationExpressionSyntax invocation && invocation.Expression == node)
+			{
+				// we already visit this node by VisitInvocationExpression, so we just skip it here
+				base.VisitMemberAccessExpression(node);
+			}
+			else
+			{
+				VisitPropertyOrIndexerAccessExpression(node);
+				base.VisitMemberAccessExpression(node);
+			}
 		}
 
 		public override void VisitElementAccessExpression(ElementAccessExpressionSyntax node)
--- a/src/Acuminator/Acuminator.Utilities/Roslyn/NestedInvocationWalker.cs
+++ b/src/Acuminator/Acuminator.Utilities/Roslyn/NestedInvocationWalker.cs
@@ -53,6 +53,8 @@ namespace Acuminator.Utilities.Roslyn
 
 		private readonly ISet<(SyntaxNode, DiagnosticDescriptor)> _reportedDiagnostics = new HashSet<(SyntaxNode, DiagnosticDescriptor)>();
 
+		private readonly SymbolInfoCache _cache = new();
+
         /// <summary>
         /// Cancellation token
         /// </summary>
@@ -119,20 +121,22 @@ namespace Acuminator.Utilities.Roslyn
 		protected virtual T? GetSymbol<T>(ExpressionSyntax node)
 			where T : class, ISymbol
 		{
-			var semanticModel = GetSemanticModel(node.SyntaxTree);
-
-			if (semanticModel != null)
+			SymbolInfo? cached = _cache.GetOrCreate(node, () =>
 			{
-				var symbolInfo = semanticModel.GetSymbolInfo(node, CancellationToken);
+				SemanticModel? semanticModel = GetSemanticModel(node.SyntaxTree);
+				return semanticModel?.GetSymbolInfo(node, CancellationToken);
+			});
 
-				if (symbolInfo.Symbol is T symbol)
+			if (cached is not null)
+			{
+				if (cached.Value.Symbol is T symbol)
 				{
 					return symbol;
 				}
 
-				if (!symbolInfo.CandidateSymbols.IsEmpty)
+				if (!cached.Value.CandidateSymbols.IsEmpty)
 				{
-					return symbolInfo.CandidateSymbols.OfType<T>().FirstOrDefault();
+					return cached.Value.CandidateSymbols.OfType<T>().FirstOrDefault();
 				}
 			}
 
--- a/src/Acuminator/Acuminator.Utilities/Roslyn/NestedInvocationWalker.cs
+++ b/src/Acuminator/Acuminator.Utilities/Roslyn/NestedInvocationWalker.cs
@@ -53,7 +53,7 @@ namespace Acuminator.Utilities.Roslyn
 
 		private readonly ISet<(SyntaxNode, DiagnosticDescriptor)> _reportedDiagnostics = new HashSet<(SyntaxNode, DiagnosticDescriptor)>();
 
-		private readonly SymbolInfoCache _cache = new();
+		private readonly SymbolInfoCache _cache;
 
         /// <summary>
         /// Cancellation token
@@ -89,6 +89,8 @@ namespace Acuminator.Utilities.Roslyn
 
 			//Use lazy to avoid calling virtual methods inside the constructor
 			_typesToBypass = new Lazy<HashSet<INamedTypeSymbol>>(valueFactory: GetTypesToBypass, isThreadSafe: false);
+
+			_cache = new SymbolInfoCache(GetType());
 		}
 
 		/// <summary>
--- a/src/Acuminator/Acuminator.Utilities/Roslyn/NestedInvocationWalker.cs
+++ b/src/Acuminator/Acuminator.Utilities/Roslyn/NestedInvocationWalker.cs
@@ -53,7 +53,7 @@ namespace Acuminator.Utilities.Roslyn
 
 		private readonly ISet<(SyntaxNode, DiagnosticDescriptor)> _reportedDiagnostics = new HashSet<(SyntaxNode, DiagnosticDescriptor)>();
 
-		private readonly SymbolInfoCache _cache;
+		private readonly SymbolInfoCache _symbolsCache;
 
         /// <summary>
         /// Cancellation token
@@ -90,7 +90,7 @@ namespace Acuminator.Utilities.Roslyn
 			//Use lazy to avoid calling virtual methods inside the constructor
 			_typesToBypass = new Lazy<HashSet<INamedTypeSymbol>>(valueFactory: GetTypesToBypass, isThreadSafe: false);
 
-			_cache = new SymbolInfoCache(GetType());
+			_symbolsCache = new SymbolInfoCache();
 		}
 
 		/// <summary>
@@ -123,7 +123,7 @@ namespace Acuminator.Utilities.Roslyn
 		protected virtual T? GetSymbol<T>(ExpressionSyntax node)
 			where T : class, ISymbol
 		{
-			SymbolInfo? cached = _cache.GetOrCreate(node, () =>
+			SymbolInfo? cached = _symbolsCache.GetOrCreate(node, () =>
 			{
 				SemanticModel? semanticModel = GetSemanticModel(node.SyntaxTree);
 				return semanticModel?.GetSymbolInfo(node, CancellationToken);
--- a/src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
+++ b/src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
@@ -1,6 +1,5 @@
 using System;
 using System.Collections.Generic;
-using System.Threading;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 
@@ -11,135 +10,22 @@ namespace Acuminator.Utilities.Roslyn;
 /// </summary>
 public sealed class SymbolInfoCache
 {
-	private readonly Dictionary<ExpressionSyntax, SymbolInfo> _map = new();
-#if SYMBOL_INFO_CACHE_STATISTICS
-	private readonly SymbolInfoCacheStatistics _statistics;
-#endif
-
-	public SymbolInfoCache(Type owner)
-	{
-#if SYMBOL_INFO_CACHE_STATISTICS
-		_statistics = SymbolInfoCacheStatisticsContext.Register(owner);
-#endif
-	}
+	private readonly Dictionary<ExpressionSyntax, SymbolInfo?> _map = new();
 
 	/// <summary>
 	/// Gets the cached symbol information for the specified expression, or creates and stores it using the provided factory.
 	/// </summary>
 	public SymbolInfo? GetOrCreate(ExpressionSyntax key, Func<SymbolInfo?> factory)
 	{
-		if (_map.TryGetValue(key, out SymbolInfo cached))
+		if (_map.TryGetValue(key, out SymbolInfo? cached))
 		{
-#if SYMBOL_INFO_CACHE_STATISTICS
-			_statistics.Hit();
-#endif
 			return cached;
 		}
 
-#if SYMBOL_INFO_CACHE_STATISTICS
-		_statistics.Miss();
-#endif
-
 		SymbolInfo? potentialValue = factory();
-		if (potentialValue is not null)
-		{
-			_map[key] = potentialValue.Value;
-#if SYMBOL_INFO_CACHE_STATISTICS
-			_statistics.Set(key.GetType());
-#endif
-			return potentialValue;
-		}
+		_map[key] = potentialValue;
 
-		return null;
+		return potentialValue;
 	}
-
-#if SYMBOL_INFO_CACHE_STATISTICS
-	public sealed class SymbolInfoCacheStatistics
-	{
-		private readonly Dictionary<Type, long> _byTypeCounter = new();
-		private readonly Type _owner;
-		private readonly int _cacheId;
-
-		private long _hitsTotal;
-		private long _missesTotal;
-		private long _setsTotal;
-
-		public SymbolInfoCacheStatistics(int cacheId, Type owner)
-		{
-			_owner = owner;
-			_cacheId = cacheId;
-		}
-
-		public void Hit() => _hitsTotal++;
-
-		public void Miss() => _missesTotal++;
-
-		public void Set(Type type)
-		{
-			if (!_byTypeCounter.ContainsKey(type))
-			{
-				_byTypeCounter[type] = 1;
-			}
-			else
-			{
-				_byTypeCounter[type]++;
-			}
-			_setsTotal++;
-		}
-
-		public StatisticsSnapshot GetSnapshot()
-		{
-			var lookupsTotal = _hitsTotal + _missesTotal;
-			return new StatisticsSnapshot(
-				_cacheId,
-				_owner,
-				_hitsTotal,
-				_missesTotal,
-				_setsTotal,
-				_byTypeCounter,
-				CalcRatio(_hitsTotal, lookupsTotal),
-				CalcRatio(_missesTotal, lookupsTotal)
-			);
-		}
-		
-		private static double CalcRatio(long value, long total) => total == 0 ? 0 : (double)value / total;
-	}
-
-	public static class SymbolInfoCacheStatisticsContext
-	{
-		private static readonly object Gate = new();
-		private static readonly List<SymbolInfoCacheStatistics> Statistics = new();
-		private static int _nextCacheId;
-
-		internal static SymbolInfoCacheStatistics Register(Type owner)
-		{
-			var cacheId = Interlocked.Increment(ref _nextCacheId);
-			var statistics = new SymbolInfoCacheStatistics(cacheId, owner);
-
-			lock (Gate)
-			{
-				Statistics.Add(statistics);
-			}
-
-			return statistics;
-		}
-
-		public static StatisticsSnapshot[] GetStatistics()
-		{
-			lock (Gate)
-			{
-				var result = new StatisticsSnapshot[Statistics.Count];
-				for (var i = 0; i < Statistics.Count; i++)
-				{
-					result[i] = Statistics[i].GetSnapshot();
-				}
-				
-				return result;
-			}
-		}
-	}
-
-	public record StatisticsSnapshot(int CacheId, Type Owner, long HitsTotal, long MissesTotal, long SetsTotal, Dictionary<Type, long> ByTypeCounter, double HitRatio, double MissRatio);
-#endif
 }
 
--- a/src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
+++ b/src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
@@ -1,0 +1,34 @@
+using System;
+using System.Collections.Generic;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace Acuminator.Utilities.Roslyn;
+
+/// <summary>
+/// Caches Roslyn symbol lookup results for expression syntax nodes visited by a syntax walker.
+/// </summary>
+internal sealed class SymbolInfoCache
+{
+	private readonly Dictionary<ExpressionSyntax, SymbolInfo> _map = new();
+
+	/// <summary>
+	/// Gets the cached symbol information for the specified expression, or creates and stores it using the provided factory.
+	/// </summary>
+	public SymbolInfo? GetOrCreate(ExpressionSyntax key, Func<SymbolInfo?> factory)
+	{
+		if (_map.TryGetValue(key, out SymbolInfo cached))
+		{
+			return cached;
+		}
+
+		SymbolInfo? potentialValue = factory();
+		if (potentialValue is not null)
+		{
+			_map[key] = potentialValue.Value;
+			return potentialValue;
+		}
+
+		return null;
+	}
+}
--- a/src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
+++ b/src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Threading;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 
@@ -8,9 +9,19 @@ namespace Acuminator.Utilities.Roslyn;
 /// <summary>
 /// Caches Roslyn symbol lookup results for expression syntax nodes visited by a syntax walker.
 /// </summary>
-internal sealed class SymbolInfoCache
+public sealed class SymbolInfoCache
 {
 	private readonly Dictionary<ExpressionSyntax, SymbolInfo> _map = new();
+#if SYMBOL_INFO_CACHE_STATISTICS
+	private readonly SymbolInfoCacheStatistics _statistics;
+#endif
+
+	public SymbolInfoCache(Type owner)
+	{
+#if SYMBOL_INFO_CACHE_STATISTICS
+		_statistics = SymbolInfoCacheStatisticsContext.Register(owner);
+#endif
+	}
 
 	/// <summary>
 	/// Gets the cached symbol information for the specified expression, or creates and stores it using the provided factory.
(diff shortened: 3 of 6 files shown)


src/.idea/.idea.Acuminator/.idea/vcs.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="VcsDirectoryMappings">
+    <mapping directory="$PROJECT_DIR$/.." vcs="Git" />
+  </component>
+</project>
```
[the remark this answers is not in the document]
SENya1990: @gelverpl  this seems like IDE specific. I think it's better to update the .gitignore file in the repo to not track the entire `.idea` folder.

src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
```diff
@@ -0,0 +1,145 @@
...
+
+namespace Acuminator.Utilities.Roslyn;
+
+/// <summary>
+/// Caches Roslyn symbol lookup results for expression syntax nodes visited by a syntax walker.
+/// </summary>
+public sealed class SymbolInfoCache
+{
+	private readonly Dictionary<ExpressionSyntax, SymbolInfo> _map = new();
+#if SYMBOL_INFO_CACHE_STATISTICS
```
SENya1990: @gelverpl we spoke today about using some kind of feature flags for performance profiling features. Did you decide not to use it in the end?

src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
```diff
@@ -0,0 +1,145 @@
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace Acuminator.Utilities.Roslyn;
+
+/// <summary>
+/// Caches Roslyn symbol lookup results for expression syntax nodes visited by a syntax walker.
```
SENya1990, line 9: IMHO, this can be already a bit generalized to be lookup for syntax nodes, not just expression syntax nodes. There are many other types of nodes that can be passed to the semantic model:
https://learn.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.semanticmodel?view=roslyn-dotnet-5.0.0

src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
```diff
@@ -0,0 +1,145 @@
...
+				_byTypeCounter,
+				CalcRatio(_hitsTotal, lookupsTotal),
+				CalcRatio(_missesTotal, lookupsTotal)
+			);
+		}
+		
+		private static double CalcRatio(long value, long total) => total == 0 ? 0 : (double)value / total;
+	}
+
+	public static class SymbolInfoCacheStatisticsContext
```
SENya1990: @gelverpl please use 1 type per file convention (except maybe 1 line records types) and move this to a separate file.

src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
```diff
@@ -0,0 +1,145 @@
...
+	{
+		private static readonly object Gate = new();
```
SENya1990: For private fields the current convention is to use underscore-prefixed camel case.

Also it's not very important, but the established naming convention for these objects is to call them `_locker`

src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
```diff
@@ -0,0 +1,145 @@
...
+			_statistics.Set(key.GetType());
+#endif
+			return potentialValue;
+		}
+
+		return null;
+	}
+
+#if SYMBOL_INFO_CACHE_STATISTICS
+	public sealed class SymbolInfoCacheStatistics
```
SENya1990: Should be moved to a separate file. 
The type should probably be marked as `internal` instead of `public`.

src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
```diff
@@ -0,0 +1,145 @@
...
+				{
+					result[i] = Statistics[i].GetSnapshot();
+				}
+				
+				return result;
+			}
+		}
+	}
+
+	public record StatisticsSnapshot(int CacheId, Type Owner, long HitsTotal, long MissesTotal, long SetsTotal, Dictionary<Type, long> ByTypeCounter, double HitRatio, double MissRatio);
```
SENya1990: Imho, this DTO can also be in a separate file, but I won't mind if it will be placed in a file with the `SymbolInfoCacheStatistics` type.

And it also should be `internal`.

src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
```diff
@@ -0,0 +1,145 @@
...
+			{
+				Statistics.Add(statistics);
+			}
+
+			return statistics;
+		}
+
+		public static StatisticsSnapshot[] GetStatistics()
+		{
+			lock (Gate)
```
SENya1990: This is not a big improvement, since `GetSnapshot` doesn't do a lot currently, but you can reduce the time under the lock in this algorithm, if you make under the lock a copy of the `Statistics` list. Then you can call `GetSnapshot` on the copy of the list without locking the `Statistics` collection.

src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
```diff
@@ -0,0 +1,145 @@
...
+			return new StatisticsSnapshot(
+				_cacheId,
+				_owner,
+				_hitsTotal,
+				_missesTotal,
+				_setsTotal,
+				_byTypeCounter,
+				CalcRatio(_hitsTotal, lookupsTotal),
+				CalcRatio(_missesTotal, lookupsTotal)
+			);
```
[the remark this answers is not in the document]
SENya1990: @gelverpl  this is a good point. It's not really a snapshot if you are passing the original reference which will be mutated after the creation of the snapshot.

src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
```diff
@@ -0,0 +1,145 @@
...
+			var statistics = new SymbolInfoCacheStatistics(cacheId, owner);
+
+			lock (Gate)
+			{
+				Statistics.Add(statistics);
+			}
+
+			return statistics;
+		}
+
```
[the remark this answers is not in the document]
SENya1990: These profiling features are created only for special profiling builds and for profiling sessions only. So, I don't think it's a problem now.

src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
```diff
@@ -0,0 +1,145 @@
...
+		public static StatisticsSnapshot[] GetStatistics()
+		{
+			lock (Gate)
+			{
+				var result = new StatisticsSnapshot[Statistics.Count];
+				for (var i = 0; i < Statistics.Count; i++)
+				{
+					result[i] = Statistics[i].GetSnapshot();
```
SENya1990: Since it's a static context, it can be shared by multiple analyzers running on different threads. Which means that while the call to `GetStatistics` is executed on one thread, the other thread may update the counters for a concrete statistic. This make this collection a bit different from a full snapshot and makes the measurement a bit inaccurate - snapshot for one statistic is collected at the moment X, snapshot for the next statistic will be collected for the moment X + some small delta.

For full snapshots you would need to use sync the updates of the counters with the retrieval of statistics... which will make things more difficult for us and introduce undesirable threads synchronization. 

I think it's OK to leave it as is since we are not micro benchmarking and looking for milliseconds. But this should be a conscious decision, so I think a comment about this decision should be added here describing it.

src/Acuminator/Acuminator.Utilities/Roslyn/NestedInvocationWalker.cs
```diff
@@ -53,6 +53,8 @@ public abstract class NestedInvocationWalker : CSharpSyntaxWalker
 
 		private readonly ISet<(SyntaxNode, DiagnosticDescriptor)> _reportedDiagnostics = new HashSet<(SyntaxNode, DiagnosticDescriptor)>();
 
+		private readonly SymbolInfoCache _cache;
```
SENya1990: There are other caches in this type, so IMHO a more specific name should be used like `_symbolsCache`.

src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
```diff
@@ -0,0 +1,145 @@
...
+		private static double CalcRatio(long value, long total) => total == 0 ? 0 : (double)value / total;
+	}
+
+	public static class SymbolInfoCacheStatisticsContext
+	{
+		private static readonly object Gate = new();
+		private static readonly List<SymbolInfoCacheStatistics> Statistics = new();
+		private static int _nextCacheId;
+
+		internal static SymbolInfoCacheStatistics Register(Type owner)
```
SENya1990: @gelverpl there can be multiple statistics registered for the same owner (walker). Is it okay to not aggregate them into one statistic?

src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
```diff
@@ -0,0 +1,145 @@
...
+			if (!_byTypeCounter.ContainsKey(type))
+			{
+				_byTypeCounter[type] = 1;
+			}
+			else
+			{
+				_byTypeCounter[type]++;
+			}
+			_setsTotal++;
+		}
```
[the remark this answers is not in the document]
SENya1990: This shouldn't be a problem for now since each instance of symbol cache and related statistic is owned by only one dedicated nested invocation walker which is executed only by one thread. 

Later, when this cache will be expanded to be shared with all analyzers, this indeed will be a problem.

src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
```diff
@@ -0,0 +1,145 @@
...
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace Acuminator.Utilities.Roslyn;
+
+/// <summary>
+/// Caches Roslyn symbol lookup results for expression syntax nodes visited by a syntax walker.
+/// </summary>
+public sealed class SymbolInfoCache
+{
+	private readonly Dictionary<ExpressionSyntax, SymbolInfo> _map = new();
```
SENya1990: @gelverpl we should cache `ISymbol?`, there is no need to keep intermediate symbol info DTO.

src/Acuminator/Acuminator.Utilities/Roslyn/NestedInvocationWalker.cs
```diff
@@ -119,20 +123,22 @@ protected virtual HashSet<INamedTypeSymbol> GetTypesToBypass() =>
 		protected virtual T? GetSymbol<T>(ExpressionSyntax node)
 			where T : class, ISymbol
 		{
-			var semanticModel = GetSemanticModel(node.SyntaxTree);
-
-			if (semanticModel != null)
+			SymbolInfo? cached = _cache.GetOrCreate(node, () =>
```
SENya1990: Here should be the retrieval of the symbol from cache, not `SymbolInfo`. 
As I wrote in https://github.com/Acumatica/Acuminator/pull/668/changes#r3197690911, cache should keep `ISymbol?`.

Also, since the symbol is not always cached, IMHO it's better to call it just `symbol`

SENya1990 (changes requested):
Overall, the PR content is good, there is no need to significantly change anything.
But I have some remarks that should be processed:
- about the type of cached data, 
- some minor remarks about the concurrent code
- remarks about the code style.

src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
```diff
@@ -0,0 +1,145 @@
...
+		private long _missesTotal;
+		private long _setsTotal;
+
+		public SymbolInfoCacheStatistics(int cacheId, Type owner)
+		{
+			_owner = owner;
+			_cacheId = cacheId;
+		}
+
+		public void Hit() => _hitsTotal++;
```
SENya1990: All counting methods should be annotated with the `System.Diagnostics.ConditionalAttribute` instead of using pragma directives at their every usage. 
https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.conditionalattribute?view=netstandard-2.0

src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	/// <summary>
+	/// Gets the cached symbol information for the specified expression, or creates and stores it using the provided factory.
+	/// </summary>
+	public SymbolInfo? GetOrCreate(ExpressionSyntax key, Func<SymbolInfo?> factory)
+	{
+		if (_map.TryGetValue(key, out SymbolInfo cached))
+		{
+#if SYMBOL_INFO_CACHE_STATISTICS
```
SENya1990: All counting methods can be annotated with the `System.Diagnostics.ConditionalAttribute` instead of using pragma directives at their every usage. 
https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.conditionalattribute?view=netstandard-2.0

It won't compile without the pragma directive as is though. So, to make it compile and get rid of to many pragmas, maybe you can always keep the field with statistics object and for regular builds initialize it to null or an empty static object.

src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
```diff
@@ -0,0 +1,145 @@
...
+			return potentialValue;
+		}
+
+		return null;
+	}
+
+#if SYMBOL_INFO_CACHE_STATISTICS
+	public sealed class SymbolInfoCacheStatistics
+	{
+		private readonly Dictionary<Type, long> _byTypeCounter = new();
```
SENya1990: Imho, the name can indicate that the keys are type **of syntax nodes**. For example, it can be `_byNodeTypeCounter`

src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
```diff
@@ -0,0 +1,145 @@
...
+		{
+			_owner = owner;
+			_cacheId = cacheId;
+		}
+
+		public void Hit() => _hitsTotal++;
+
+		public void Miss() => _missesTotal++;
+
+		public void Set(Type type)
```
SENya1990: Imho, `nodeType` is clearer

src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
```diff
@@ -0,0 +1,145 @@
...
+#endif
+			return cached;
+		}
+
+#if SYMBOL_INFO_CACHE_STATISTICS
+		_statistics.Miss();
+#endif
+
+		SymbolInfo? potentialValue = factory();
+		if (potentialValue is not null)
```
SENya1990: Why not caching null results? It's not like the semantic model will return a different result with another call.

src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
gelverpl: Agree. Fixed.

src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
gelverpl: Agree. Fixed.

src/Acuminator/Acuminator.Utilities/Roslyn/NestedInvocationWalker.cs
gelverpl: Agree. Fixed.

src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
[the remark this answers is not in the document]
gelverpl: Actually, this is not. All of the above is irrelevant, as no one collects (and should never do) these statistics except the person who requests these statistics in debug mode.
I've broken a lot of rules while working on this code, since the only situation in which this code runs is a controlled run in a debugger. Furthermore, I suspect the only "consumer" of this code will be me.
The plan is to get rid of cache hits -> get rid of statistics -> get rid of cache.

Won't be fixed.

src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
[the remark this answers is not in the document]
gelverpl: This cache is strictly reserved for its intended code and **must not** be accessed or reused elsewhere.

**When** and **if** we decide to implement some kind of shared cache, it will be a completely new and unrelated implementation.

Won't be fixed.

src/.idea/.idea.Acuminator/.idea/vcs.xml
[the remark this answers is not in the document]
gelverpl: Among JetBrains users, it is generally accepted practice to version-control only those XML files within .idea that define shared project rules and coding standards, while ignoring everything related to a developer's personal workspace.
If it bothers you too much, I'll remove it.

src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
gelverpl: I was more talking about enabling/disabling certain optimizations to make the measurement process more convenient. As for these statistics, I think the only person who will need them is me. So I don't see any need for anything more sophisticated here.

src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
gelverpl: The only user of this cache is `GetSymbol<T>(ExpressionSyntax node)`, And no one else should use it.
Let me point out that the critical issue I fixed alongside introducing the cache in the `VisitMemberAccessExpression` method was caused by over-generalization. I disagree with your point.

Won't be fixed.

src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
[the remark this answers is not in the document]
SENya1990: >All of the above is irrelevant, as no one collects (and should never do) these statistics except the person who requests these statistics in debug mode.

@gelverpl you can't guarantee that or predict who will work on this code in two-three years.

> I've broken a lot of rules while working on this code, since the only situation in which this code runs is a controlled run in a debugger.  Furthermore, I suspect the only "consumer" of this code will be me.

Again:
1. The support and work on this code only by yourself is not guaranteed, at least from my POV. This is the first contribution and with all due respect for the contribution, as a maintainer I just don't have enough info to predict whether there will be other contributions. There are things like the bus factor too. So, I would prefer to keep the code consistent. Considering that making a copy of a dictionary is simple, saying that somewhere in the future it will be removed IMHO is not enough to justify this.

2. I wouldn't be so sure that the symbol cache will be deleted. IMHO, it can be reworked and extended, so it will be reused by all analyzers which access the same syntax tree and call `GetSymbolInfo` on its nodes. In that case this no longer will be a situation when some analyzer redundantly checks the same node twice.

3. In general, "only I will work on this code" is not a good justification for breaking the rules. Especially, if the fix is quite simple and there is no urgent deadline.

4. Nothing says that these statistics should be used only in the debug mode (a comment will be good). 
And even if you use them in debug mode, you still won't have a correct snapshot for `_byTypeCounter` in the general case.  Suppose, that you  want to compare statistics during the debug between two snapshots, 1 and 2 and see the difference in the collected `_byTypeCounter` values. You won't be able to see it since both snapshots keep the reference to the same dictionary.

src/.idea/.idea.Acuminator/.idea/vcs.xml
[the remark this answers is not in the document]
SENya1990: No need, the files can stay. I just didn't know that JB recommends to share some files under its IDE related folder.

src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
SENya1990: I disagree with the part about the only person using them. It's hard to predict.

However, this was just a question about the previous discussion. I'm ok with this approach.

src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
SENya1990: Ok, let's keep it like this for now, it may be an over-generalization indeed.

src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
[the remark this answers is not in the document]
gelverpl: Fixed

src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
gelverpl: Fixed.

src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
gelverpl: Fixed. You should mention this in your conventions, I mean specifically static fields and _locker for sync obj

src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
gelverpl: Fixed

src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
gelverpl: Fixed.

src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
gelverpl: No it wasn't possible 1 walker -> 1 cache instance -> 1 registered statistics.
Fixed.

src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
gelverpl: I find this type more convenient for analysis.

Won't be fixed.

src/Acuminator/Acuminator.Utilities/Roslyn/NestedInvocationWalker.cs
gelverpl: I find this type more convenient for analysis.

Won't be fixed.

src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
gelverpl: Fixed.

src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
gelverpl: Good point. Тo longer relevant. Fixed.

src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
gelverpl: Fixed.

src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
gelverpl: Fixed.

src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
gelverpl: Fixed.

src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
SENya1990: Agree. Due to not many contributors many coding conventions weren't fixed explicitly. I will add a section here
https://github.com/Acumatica/Acuminator/blob/dev/docs/dev/CodingGuidelines/CodingGuidelines.md

src/Acuminator/Acuminator.Utilities/Roslyn/NestedInvocationWalker.cs
SENya1990: Could you please elaborate why is it more convenient for analysis?

src/Acuminator/Acuminator.Utilities/Roslyn/SymbolInfoCache.cs
SENya1990: > No it wasn't possible 1 walker -> 1 cache instance -> 1 registered statistics.

I meant the aggregation by walker type, since the owner is a `Type`, not an instance of `NestedInvocationWalker`.
If the statistics was intended to be measured per walker instance, then all is good (though a bit strange).

If it was intended to be measured by walker type, then it seems that the aggregation is still needed. In general, an analyzer can run a registered action on multiple tree nodes or symbols in the file, so the same type of walker can be created more than once for a single syntax tree, If multiple files are analyzed, then there will be at least one walker per syntax tree,

This can be seen in the debug of the collected statistic:
image

I just wanted to clarify an incorrect assumption because it looks like statistic per walker. Since you decided to completely remove the instrumentation (which IMHO is a pity), this doesn't matter for the PR.

SENya1990 (approved):
It's a pity that the instrumentation was removed instead of being a bit polished.
However, with its removal most of my remarks is gone.

There is really no reason to cache `SymbolInfo` instead of `ISymbol` - in all scenarios with the compiling code there should be only `Symbol` and no `CandidateSymbols`. But I can live with this.

And I will need to update the dev guidelines page.

